PR #40, opened 2014-12-05
Add method to generate a webhook signature.

### LINKED ISSUE #39: Add method to generate a webhook signature
This method would be used to validate the webhook signature is valid.

Based on the documentation:
> The signature is calculated as follows:     
> SHA1-Hex-Digest("{data}{timestamp}{api_secret}")

Usage:
````
    var data = response.body;
    var x_ts = response.header['X-Cld-Timestamp'];
    var x_sig = response.header['X-Cld-Signature'];

    var sig = cloudinary.utils.webhook_signature(data, x_ts)
    
    if (sig == x_sig) {
        console.log('valid');
    } else {
        console.log('invalid');
    }
````

donaldjarmstrong:
This is used to validate that the webhook API call is legit.  Closed #39 
Added a single test

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -378,6 +378,15 @@ exports.sign_request = (params, options = {}) ->
 
   return params
 
+exports.webhook_signature = (data, timestamp, options = {}) ->
+  throw "Must supply data"  unless data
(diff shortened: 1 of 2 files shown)


FREEZX:
Why is this not merged yet?

PR #54, opened 2015-05-19
Update cloudinary.js to copy over options instead of mutating

gching:
Instead of using in the passed in `options` argument in `.image`, we use `_.extend` to make a shallow copy of the `options` object passed in. Therefore, there will be no way of mutating the `options` object directly.

This happens in `.video` too but the `options` are deep cloned later on. I still used `.extend` as I do not know if anything before the deep clone mutates `options`.

Added tests as well :)

Issue: #55 

Side note: We are not deep copying so deeper objects could possibly be mutated. Not worrying about this unless objects to get mutated at a deeper level.

### DIFF
--- a/cloudinary.js
+++ b/cloudinary.js
@@ -12,7 +12,7 @@ exports.url = function(public_id, options) {
 
 exports.image = function (source, options) {
   var responsive, html, current_class, classes;
-  options = options || {};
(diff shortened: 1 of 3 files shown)


tocker:
Thanks for catching this!

PR #58, opened 2015-08-08
Allow to pass cloudinary url in config method

ertrzyiks:
I think it would be nice to explicitly pass clodinary_url to config method. Im not 100% happy with this commit since signature with single string param is already taken by getter. Added some tests to cover old and new use cases.

What do you think?

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -1,25 +1,32 @@
 _ = require("lodash")
 cloudinary_config = undefined
+
+parseConnectionUrl = (url) ->
+  uri = require('url').parse(url, true)
+  cloudinary_config =
+    cloud_name: uri.host,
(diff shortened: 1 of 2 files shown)


tocker:
Hi @ertrzyiks 
Sorry for not responding sooner!
We prefer not to overload the config() method with different semantics.
Since you are calling the method explicitly, why no just set the values?

For example:

```javascript
cloudinary.config({ 
  cloud_name: "demo", 
  api_key: "somekey"
});
```

ertrzyiks:
Hi @tocker,
I think I did that PR and used fork for easier intergration with some grunt or gulp plugin. I don't like idea of overloading method neither. Thanks for answer anyway!

PR #66, opened 2015-09-16
fixes #65 - upload_large using chunk_size is corrupting data

### LINKED ISSUE #65: upload_large using chunk_size is corrupting data
I'm trying to upload an 8mb file ( type raw or image ) using a chunk_size of 6000000. What I am seeing is that on completion, the file size matches, but the file data is corrupted around the location of the second chunk.

I've tried several different files including dummy test files, jpegs, and pdfs of different sizes. If the file is transmitted using a single chunk ( chunk_size is larger than file size ) , the md5 matches. If multiple chunks are involved, the md5 is different, and the md5 changes as you experiment with different chunk sizes ( 5500000 vs 6000000 for example )

Can you update your tests for upload_large ( and all other upload tests? ) to check filesize AND compare etag/md5 to the original md5 to verify the uploaded data matches.

bompus:
fixes #65 - upload_large using chunk_size is corrupting data - also adds the very useful upload_large_stream function. upload_large tests now verify data integrity.


owenallenaz:
+1 on this.

bompus:
Is there an ETA on integrating this?

PR #83, opened 2016-01-25
Migrate depreciated _.include to _.includes

gvsboy:
The lodash _.include method is depreciated and is removed in 4.0.0. This small PR updates the usage to _.includes, which is already being used by uploader.js

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -67,7 +67,7 @@
     }
     handle_response = function(res) {
       var buffer, err_obj, error;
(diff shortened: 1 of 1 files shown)


tocker:
Hi @gvsboy,

Thanks for the pull request!
You submitted the `js` file instead of the `coffee` source, so I updated the file directly instead of merging.

gvsboy:
Ack--sorry about that. Rookie mistake. Thanks for fixing this!

PR #86, opened 2016-02-01
optimized speed of generate_transformation_string

bompus:
Benchmark: https://gist.github.com/bompus/c439ec71844ff1cbea26

My results were.. 
before 2034ms ( [profile screenshot](http://screencast.com/t/vAkBYhKkodE) )
after 256ms ( [profile screenshot](http://screencast.com/t/9WHGps1a0F) )

While it might not seem to do much on a single run, it does when you have a heavy traffic, image heavy site. We were actually seeing this function call jump to the top of our v8 profiling regularly due to the synchronous code it runs. lodash.sortBy may be powerful, but unnecessary in this case.


bompus:
messed PR up.. closing for a bit.

tocker:
@bompus thanks for giving it a go!
Let us know when you got it right :smiley:

PR #87, opened 2016-02-01
optimized speed of generate_transformation_string

bompus:
Benchmark: https://gist.github.com/bompus/c439ec71844ff1cbea26

My results were.. 
before 2034ms ( [profile screenshot](http://screencast.com/t/vAkBYhKkodE) )
after 256ms ( [profile screenshot](http://screencast.com/t/9WHGps1a0F) )

While it might not seem to do much on a single run, it does when you have a heavy traffic, image heavy site. We were actually seeing this function call jump to the top of our v8 profiling regularly due to the synchronous code it runs. lodash.sortBy may be powerful, but unnecessary in this case.


bompus:
Do you have an ETA on when this can be pulled in? It's really impacting our server-side performance of image-heavy pages and high traffic sites, since we utilize Cloudinary for nearly every image.

tocker:
@bompus thanks for the pull request!
It looks great - except that you included the `js` and `map` files which should not be there.
Can you please update the PR so that only the coffee file is included?

I'll update the `.gitignore` file to explicitly exclude `js` and `map`, after merging this.

Thanks!

bompus:
@tocker pull request updated to remove js and map files.

Thank you!

tocker:
@bompus thanks for updating the PR!

bompus:
@tocker Do you have an ETA on when the release containing this fix will be made, and published to npm?

PR #111, opened 2016-09-13
escape commas only once when uploading an image with text overlay sin…

eitanp461:
…ce it is an eager transformation


src/utils.coffee
```diff
@@ -86,7 +86,7 @@ process_if = (ifValue)->
 # Parse layer options
 # @return [string] layer transformation string
 # @private
-process_layer = (layer)->
+process_layer = (layer, escapeQuotes = true)->
```
shedar: Can you suggest a better param name than `escapeQuotes`?
As I see, it's about comma and slash, not about quotes.

src/utils.coffee
```diff
@@ -271,7 +273,7 @@ exports.generate_transformation_string = (options) ->
   if options["offset"]?
     [options["start_offset"], options["end_offset"]] = split_range(utils.option_consume(options, "offset"))
 
-  overlay = process_layer(utils.option_consume(options, "overlay"))
+  overlay = process_layer(utils.option_consume(options, "overlay"), escapeQuotes)
```
shedar: Shouldn't it be added to the underlay as well?

test/utils_spec.coffee
```diff
@@ -385,6 +385,18 @@ describe "utils", ->
...
+        overlay: {
+          text:"Hello,World", 
+          font_family:"arial", 
+          font_size: '30'   
+        }
+    }
+    expect(utils.build_upload_params(options)['transformation']).to.eql("l_text:arial_30:Hello%2CWorld")
+    expect(Object.keys(options).length).to.eql(1)
+
+
```
shedar: Is test of double escaping outside of `build_upload_params` intentionally omited to be tested in core or just missing?

src/utils.coffee
asisayag2: @shedar , how about "shouldEscape"?

src/utils.coffee
asisayag2: @shedar , we should

test/utils_spec.coffee
asisayag2: @shedar missing

shedar:
Replaced by #249

PR #148, opened 2017-05-04
Feature/misc and params completion

nadavofi:
- add expires_at parameter to archive_params
- test transformations api with next_cursor
- add format parameter support to responsive-breakpoints encoder


test/api_spec.coffee
```diff
@@ -389,9 +391,14 @@ describe "api", ->
       cloudinary.v2.api.transformations (error, result) ->
         return done(new Error error.message) if error?
         transformation = find_by_attr(result.transformations, "name", EXPLICIT_TRANSFORMATION_NAME)
+        expect(result.next_cursor).not.to.be.empty()
         expect(transformation).not.to.eql(undefined)
         expect(transformation.used).to.be.ok
-        done()
+        previous_cursor = result.next_cursor
```
tocker: This is not needed since we are verifying cursor support in https://github.com/cloudinary/cloudinary_npm/blob/ea7f196b73c6c3f2a9d5fa32b06b0fdae5ddaddf/test/api_spec.coffee#L387

Issue #73, opened 2015-12-16
[Suggestion] Replace CoffeeScript with ES6


runlevel5:
Most of convenient ways that CoffeeScript offers are now available in ES6. Plus, standard ES is the way that the JS community is moving forward.

tocker:
:smile: Always a hot button issue...
There are pros and cons for each tool. What CoffeeScript has going for it is the similarity to Ruby syntax and the adoption in the Rails community. Comprehensions and the existential operator are nice too.

We're keeping a close watch on this issue and may migrate in the future.

joneslee85:
@tocker from a Rails-y perspective, yes, I wouldn't mind starting a new JS-based project with CoffeeScript. But taking into the fact this is an npm module that is targeting JS community, I believe that using ES5 or ES2015 (meh ES6) is a better choice as it is more familiar and standardised, thus could lead to more contribution.

Flo-Schield-Bobby:
+1

verlok:
+1

raiyankamal:
+1

Philin-Anton:
Hello everybody.
I would look towards TypeScript.

flo-sch:
@Philin-Anton I'm curious to know what would makes you prefer TypeScript over ES6 ?

Philin-Anton:
@flo-sch  Hi, strict typing, more explicit code division. But I thought the best option for this project would be js-flow (https://flow.org/)
In any case, you need to create a basis for translating to JS.
What do you think of it?

roeeba:
Closing this issue due to the time elapsed. Please feel free to either re-open the issue, contact our support at http://support.cloudinary.com or create a new ticket if you have any additional issues.

PR #155, opened 2017-08-11
#73 [Suggestion] Replace CoffeeScript with ES6

Philin-Anton:
Created a bass for seamless transition to ES6


Philin-Anton:
Since I did not touch the sources and tests.
And I could not break down them ...
Would you be able to fix the tests?

PR #160, opened 2017-08-18
Support of "transformations" parameter in Delete Resources

RTLcoil:


test/api_spec.coffee
```diff
@@ -711,7 +711,7 @@ describe "api", ->
   # Replace `it` with  `it.skip` below if you want to disable it.
   it "should list folders in cloudinary", (done)->
     @timeout helper.TIMEOUT_LONG
-    Q.all([
+    return Q.all([
```
tocker: Why return?

test/api_spec.coffee
```diff
@@ -732,8 +732,8 @@ describe "api", ->
       expect(sub_1.folders[1].path).to.eql('test_folder1/test_subfolder2')
       cloudinary.v2.api.sub_folders('test_folder_not_exists')
     ).then((result)->
-      console.log('error test_folder_not_exists should not pass to "then" handler but "catch"')
```
tocker: Why remove the message?

src/api.coffee
```diff
@@ -153,24 +153,24 @@ exports.delete_resources = (public_ids, callback, options = {}) ->
   resource_type = options["resource_type"] ? "image"
   type = options["type"] ? "upload"
   uri = ["resources", resource_type, type]
-  call_api("delete", uri, _.extend({"public_ids[]": public_ids}, api.only(options, "keep_original", "invalidate")), callback, options)
+  call_api("delete", uri, _.extend({"public_ids[]": public_ids}, api.only(options, "keep_original", "invalidate", "transformations")), callback, options)
```
tocker: `transformations` should be an array serialised as eager transformation. I don't see it here.

Also, Please add tests.

package.json
```diff
@@ -27,6 +27,7 @@
     "pretest": "npm run compile",
     "prepublish": "npm run compile",
     "test": "node_modules/.bin/mocha -R spec --recursive test/",
+    "test:watch": "node_modules/.bin/mocha -R spec --watch test/",
```
tocker: Unrelated to the task. Please put in a separate commit.

test/api_spec.coffee
RTLcoil: Seems like a line was removed during the merge but we don't know why (the developer doesn't remember...)

test/api_spec.coffee
RTLcoil: To get the test working.

src/api.coffee
RTLcoil: Transformations are formatted on caller side and just retransferred to Clodinary there, so no data transformations required. We'll check how UT could be added for this case.

package.json
RTLcoil: Sure, we'll remove it.

RTLcoil:
Please review the following PR with your comments implemented: https://github.com/cloudinary/cloudinary_npm/pull/178

PR #164, opened 2017-09-04
Support quality override using the update resource Admin API and 'explicit' Upload API (NodeJS)

RTLcoil:


src/utils.coffee
```diff
@@ -180,6 +180,7 @@ process_layer = (layer)->
 
 exports.build_upload_params = (options) ->
   params =
+    quality_override: utils.as_safe_bool(options.quality_override)
```
tocker: The type of the parameter is string (like quality).

The list of parameters is sorted. This should be in line 200

test/api_spec.coffee
```diff
@@ -484,7 +486,11 @@ describe "api", ->
         true
       validate_presets = ->
         cloudinary.v2.api.upload_presets (error, response) ->
-          expect(response.presets.slice(0,3).map((p) -> p.name)).to.eql(delete_names)
+          count = delete_names.length
+          for preset in response.presets
+            if _.includes(delete_names, preset.name) && count > 0
+              count = count - 1;
+          expect(count).to.eql(0)
```
tocker: Why?

test/api_spec.coffee
```diff
@@ -611,6 +617,20 @@ describe "api", ->
         true
       true
 
+    it "should successfully call update api with 'quality_override'", (done) ->
+      @timeout helper.TIMEOUT_LONG
+      options = {
+        quality_override: 'auto:best',
+        moderation_status: "approved"
```
tocker: Why is this test mixed with moderation?

RTLcoil: We just based it on other examples we've seen here and there, to test with other functionality, many of your tests are written like this.

test/api_spec.coffee
RTLcoil: Otherwise, tests do not pass. Not always, but it did happen. The reason was that this code didn't delete everything after test's run.

src/utils.coffee
RTLcoil: Sure, we'll change it.

PR #176, opened 2017-12-20
adding filename option for streaming uploads

abachuk:
when streaming a file, it's still important to be able to pass the filename as an option 
```
cloudinary.uploader.unsigned_upload_stream('presetName', (result) => {}, {filename: req.file.originalname})
```

### DIFF
--- a/src/uploader.coffee
+++ b/src/uploader.coffee
@@ -276,7 +276,7 @@ call_api = (action, callback, options, get_params) ->
 post = (url, post_data, boundary, file, callback, options) ->
   finish_buffer = new Buffer("--" + boundary + "--", 'ascii')
   if file? || options.stream
-    filename = if options.stream then "file" else path.basename(file)
+    filename = if options.stream then options.filename || "file" else path.basename(file)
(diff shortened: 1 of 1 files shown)


src/uploader.coffee
```diff
@@ -276,7 +276,7 @@ call_api = (action, callback, options, get_params) ->
 post = (url, post_data, boundary, file, callback, options) ->
   finish_buffer = new Buffer("--" + boundary + "--", 'ascii')
   if file? || options.stream
-    filename = if options.stream then "file" else path.basename(file)
+    filename = if options.stream then options.filename || "file" else path.basename(file)
```
TheDancingCode: Wouldn't it be better to do 
```js
path.basename(options.filename)
```
similar to
```js
path.basename(file)
```

TheDancingCode:
@tocker 
Any thoughts on this?

src/uploader.coffee
tocker: Makes sense.

dorelljames:
So we ran into this scenario and we find this really useful to get merged, please. Thanks

shedar:
Replaced by #250

PR #192, opened 2018-04-16
add context parameter to the explicit method

oroce:
Based on the [docs](https://cloudinary.com/documentation/image_upload_api_reference#explicit) and on my recent investigation `context` is parameter should be added to the `explicit` method.


tocker:
Hi @oroce ,
Thanks for the PR!
You are correct, however `explicit` is supposed to support all of `upload`'s parameters, and the full solution was implemented in #196.

PR #199, opened 2018-07-07
fix normalize_expression when variable is names as a keyword, like "$width"

drico:
This fixes the use case when a user sets a variable named like an
keyword, like "$width".

Currently, the definition is ok ("$width_10" is untouched), but usage
is not ok ("w_$width" is converted to "w_$w") which breaks the url).


strausr:
As $width is a reserved keyword this will not be implemented at this time

PR #207, opened 2018-09-20
Add support for gs:// urls in uploader

igy:
URL handling in upload() has special cases for non-local URLs but doesn't support gs: URL prefix
Without this, attempts to use gs: URLs are handled locally, and the file can't be loaded.

Tests pass except `should support requesting ocr analysis` which appears unrelated, and was based on running tests against a cloud which hasn't got access to OCR

Verified `gs://` upload works with an example in my own account, fetched from my own storage bucket on Google Cloud (cloud `stephencloudinary`, public ID `c3uirc0l0gq0eylnhiv0`)

### DIFF
--- a/lib-es5/uploader.js
+++ b/lib-es5/uploader.js
@@ -67,7 +67,7 @@ exports.upload = function upload(file, callback) {
 
   return call_api("upload", callback, options, function () {
     var params = build_upload_params(options);
-    if (file != null && file.match(/^ftp:|^https?:|^s3:|^data:[^;]*;base64,([a-zA-Z0-9\/+\n=]+)$/)) {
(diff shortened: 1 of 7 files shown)


igy:
reclaiming and clearing review request until unit test is added

tocker:
@igy can you please rebase onto master to resolve conflicts in `lib/uploader.js`?

igy:
I'm a little confused about why the Travis build is failing in node 4 + 6 - I'll have to look into it some more

tocker:
@igy node 4 + 6 failed because you didn't run `npm run compile`. Your changes do not appear in `lib-es5`.

igy:
Yep, that was it - thanks for pointing me in the right direction; I missed that step in the build instructions and I was using ES6 so didn't notice it wasn't updated correctly

PR #239, opened 2019-03-04
Remove coffee script dependencies and transform

tornqvist:
The browserify transform defined in `package.json` was causing my build to fail due to coffeescript not being a listed dependency.

Seeing as the source code have been refactored in JavaScript, there's no need for the transform anymore. This removes any last trailing references to coffescript.


tocker:
@tornqvist thanks for the PR!

PR #248, opened 2019-03-22
Add support for 'delete folder' API

igy:
Add SDK support for the 'delete folder' API, which permits deletion of empty folders.
Add tests for the delete folder method

Example Usage:

    cloudinary.v2.api.delete_folder("temp", function(error, result){console.log(result)});

### DIFF
--- a/lib-es5/api.js
+++ b/lib-es5/api.js
@@ -458,6 +458,14 @@ exports.sub_folders = function sub_folders(path, callback) {
   return call_api("get", uri, {}, callback, options);
 };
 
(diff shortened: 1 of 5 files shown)


igy:
I'm not crazy about the delay() calls in the test, but it seems the index takes a little while to update, and without them it was failing to find the newly created folder and/or was still finding it in a call immediately after the deletion call - open to suggestions for other ways to achieve the same thing

igy:
That last commit was not correct; will fix

igy:
Rebased, squashed, amended

PR #249, opened 2019-03-27
Fix overlay/underlay commas encoding on upload

shedar:
ES6 implementation of #111


strausr:
will be implemented on future pr

PR #250, opened 2019-03-27
Add `filename` option for streaming uploads and filename overriding

shedar:
ES6 implementation of #176 with added tests.


patrick-tolosa:
Related to #273  and #274  - closing to continue discussion there

PR #255, opened 2019-04-15
Escape quotes in HTML attributes (NodeJS)

RTLcoil:
When creating an HTML tag encodes quotes inside attributes using HTML entities.

**Semantic nitpicking**: In all the SDKs this function is called `escapeQuotes` but what it actually does is encode quotes, not escape them. Renaming the functions should be relatively painless as this is an internal private function and should not affect the public API.


RTLcoil:
Replaced by #259

PR #260, opened 2019-04-22
Support different radius for each corner (containing images and overlays)

RTLcoil:
Added support for a radius option which can receive a number, a string (containing either the value or a colon separated list of values), or an array of numbers and/or strings.

Replaces #258


RTLcoil:
Broken Travis test is due to unrelated issue in npm 4. Same result can be seen here: https://travis-ci.org/cloudinary/cloudinary_npm/builds/522995139

PR #262, opened 2019-04-25
Added automatic JavaScript linting and fixed existing formatting errors

RTLcoil:
As discussed, the PR was divided into multiple commits so it is easier to grok and approve.

All the commits except the last one (5c2f766) are fixes for errors and inconsistencies discovered through linting.

The last commit contains the additions needed to let anyone lint the project by running `npm run lint`. Those using supporting code editors and IDEs (e.g., VS Code) will benefit from ESLint highlighting errors in their editor as they code.

In a future PR, linting can be added as a pre-commit hook to ensure that all commits comply with the accepted code standard, quality is maintained, less bugs are introduced, and balance is restored to the force.

You will notice that in the [ESLint config file](https://github.com/cloudinary/cloudinary_npm/blob/5c2f76627ce0fb77f07a7b6c6f9c52f79be7af94/.eslintrc.json) I have disabled many of the linting rules. This was done so as to limit the number of meaningful changes to the code that had to be introduced in this PR, and yet still allow ESLint to pass.

The rules in lines [40-73](https://github.com/cloudinary/cloudinary_npm/blob/5c2f76627ce0fb77f07a7b6c6f9c52f79be7af94/.eslintrc.json#L40) in particular are ones I believe we should enforce. For example: [no-dupe-keys](https://eslint.org/docs/rules/no-dupe-keys) has found places where the code erroneously uses the same key twice in an object definition [(utils_spec.js)](https://github.com/cloudinary/cloudinary_npm/blob/master/test/utils_spec.js#L1226), there are variables that are defined but never used, class constructors that are empty and redundant, unreachable code (usually code that comes after a return statement), and more.


RTLcoil:
Note that the breaking Travis build is due to an unrelated issue with Webpack's post install script in Node 4 🤷‍♂️

.eslintrc.json
```diff
@@ -0,0 +1,75 @@
...
+  },
+  "extends": "airbnb-base",
+  "globals": {
+    "Atomics": "readonly",
+    "SharedArrayBuffer": "readonly"
+  },
+  "parserOptions": {
+    "ecmaVersion": 2018
+  },
+  "rules": {
```
tocker: We need to discuss each overriding rule here.

Can be done after merging this PR.

tocker: Sort rules keys.

lib/utils/index.js
```diff
@@ -395,7 +394,8 @@ exports.encode_double_array = function encode_double_array(array) {
 
 exports.encode_key_value = function encode_key_value(arg) {
   if (isObject(arg)) {
-    return entries(args).map(([k,v])=>`${k}=${v}`).join('|');
+    /* eslint-disable-next-line no-undef */
+    return entries(args).map(([k, v]) => `${k}=${v}`).join('|');
```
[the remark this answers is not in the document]
RTLcoil: This was done in a separate pull request to separate the bug fix from this new feature. I am closing #263 and fixing here as you requested.

RTLcoil:
All requested changes made.

In addition, note that the new `max-len` rule requested led to some additional changes that were required and were done in a separate commit for easier grokking.

bd82:
Hello.

Some suggestions:
- **Consider using eslint only for semantic errors and something like [prettier](https://prettier.io/) for syntax formatting.**
  - Tools such as prettier/standardJS are more specialized and focused on high quality and automated formatting, but they are also very opinionated which means one needs to get used to this workflow 😄 
  - https://prettier.io/docs/en/integrating-with-linters.html#eslint
- Use [lint-staged](https://github.com/okonet/lint-staged) to **run linters/formatters ***incrementally*** on pre-commit hook.**
- **Evaluate using another (not airbnb-base) [popular  eslint configuration pre-set**](https://github.com/dustinspecker/awesome-eslint#configs) instead of **manually** managing so many eslint options in the configuration, perhaps there is a less strict one? or perhaps some of the errors can be auto-fixed
using `eslint --fix` ? Basically try to be closer to a pre-set config if possible.


A good guide for these recommendations can be found here (ignore the Jest related parts).
- https://hungryturtlecode.com/projects/eslint-husky-parallax/

Cheers.

bd82:
Also I am not sure the lint task would actually run during the Travis build, should not it be integrated to run on every build and fail it if new linting errors were introduced?

PR #263, opened 2019-04-27
Fixed bug in utils encode_key_value

RTLcoil:


RTLcoil:
Note that the breaking Travis build is due to an unrelated issue with Webpack's post install script in Node 4 🤷‍♂️

RTLcoil:
Will be fixed in #262

PR #265, opened 2019-05-14
Added support for sources in video tag

RTLcoil:
(Update video HTML 5 tag to include support for h265 and vp9)


RTLcoil:
Travis results:
* Node 4: crashes because of unrelated bug
* Node 6: Passes
* Node 8 && 10: An unrelated test timed out.

lib-es5/cloudinary.js
```diff
@@ -145,29 +146,35 @@ exports.video = function video(public_id, options) {
...
+  var multi_source_types = _.isArray(source_types) && source_types.length > 1;
+  var has_sources = _.isArray(sources) && sources.length > 0;
   var source = public_id;
-  if (!multi_source) {
+  if (!multi_source_types && !has_sources) {
     source = source + '.' + cloudinary.utils.build_array(source_types)[0];
   }
   var src = cloudinary.utils.url(source, video_options);
-  if (!multi_source) video_options.src = src;
+  if (!multi_source_types && !has_sources) video_options.src = src;
```
asisayag2: @RTLcoil  Why we need this and line 152? it's the same condition

RTLcoil: We can remove the duplication of the condition but the resulting code would be even less elegant and even longer as a result - it will look like [this diff](https://gist.github.com/RTLcoil/af95b5d4cff92a6749038c41e71e0459/revisions).

Just to be clear, this duplicate repeating condition was in the code before the pull request. It is needed because there is code that needs to run under both conditions (`let src = cloudinary.utils.url(source, video_options);`) followed by code that only runs if one of the conditions match.

At first glance I thought we could simply move `let src = cloudinary.utils.url(source, video_options);` into the condition since if that condition doesn't match `src` isn't used again outside the conditional... but that isn't possible because `cloudinary.utils.url()`is not a pure function. It modifies the state and so if it isn't called all the tests break.

PR #266, opened 2019-05-15
Removed legacy versions of nodejs.

bd82:
Node 4 and 6 have reached End Of Life:
- https://github.com/nodejs/Release#end-of-life-releases

This means that the compilation using Babel to ES5 is no longer
needed which reduces the complexity of this project.

Note that nodejs version 12 was added to the travis test suite.


bd82:
How come does the travis-ci build not execute?

strausr:
@bd82 Thank you for the contribution. This will be handled on https://github.com/cloudinary/cloudinary_npm/pull/304

PR #267, opened 2019-05-16
Add 'derived_next_resource' to api.resource method

igy:
Allows pagination when accessing the details of a single resource and many derived versions exist

Unit test to verify the parameter is passed to the server when specified


igy:
Reasonably sure the Travis build failure isn't related to these changes, so marking as ready for review

PR #268, opened 2019-05-18
TypeScript Signatures

bd82:
WIP on creating TypeScript signatures.

### DIFF
--- a/api.d.ts
+++ b/api.d.ts
@@ -1,12 +1,10 @@
 interface Common {
-    // TODO: obtain signatures for the `config` method
     config(new_config: any, new_value: any)
 
-    // TODO: is `utils` part of the official API?
-    //       if not, why is it exposed at runtime?
+    // TODO: the utils module seems to mix both internal and potentially external utils
+    //       which are an external API and which are only internal?
+    //       Could this be represented at runtime by avoiding the exposure of "internal APIs"?
     utils: any
-    // TODO: add uploader details
-    uploader: any
 
     PreloadedFile: {
         new(file_info: any): PreloadedFile
@@ -14,7 +12,6 @@ interface Common {
 
     Cache: CacheInterface
 
-    // TODO: document the URL options
     url(public_id: string, options?: any): string
 
     /**
@@ -120,7 +117,6 @@ interface Common {
 }
 
 interface V1 extends Common {
-    // TODO: are these part of the official API?
     api: API_V1
     uploader: Uploader_V1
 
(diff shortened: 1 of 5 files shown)


lib/cloudinary.js
```diff
@@ -36,22 +36,6 @@ function chainTransformations(options, transformation=[])
   return urlOptions;
 }
 
-/**
```
tocker: It would be better to keep the comments here, to support anyone looking at the source code.

bd82: This is a good point regarding assisting **developers** looking at the source code.
However the Type Signatures can also serve as documentation for **end users**.
For example: go-to-definition in VSCode would re-direct to the .d.ts file if available.

My suggestion is to
- Keep the documentation as part of the TypeScript signatures.
-  **link** the source code implementation to the TypeScript signatures
   - [Example](https://github.com/cloudinary/cloudinary_npm/pull/268/files#diff-7b97ff93a95445421f186ba62c704ca5R137)
  - Note that this `@link {...}` syntax is traversal in webstorm/intellij so one can navigate to the APIs/Docs from the source code.

This pattern would enable in the future to integrate the TypeScript compiler
as a TypeChecker on the **JavaScript** code, thus adding some automated verification
that the **implementation actually matches the interface**.
- More info on this [here](https://github.com/bd82/typescript_for_public_apis)

bd82:
All-right, I fear this is as much time as I have to invest in this topic right now.

This should provide a fairly complete skeleton for the Type Signatures based on the source code.
To road to completion would be:
- Figure out the actual types of many of the params/return types (all the **any** types).
- Figure out what part of the `utils` namespace are actually public apis.
- Figure out if there is anything else that is exposed at runtime (via require("cloudinary"))
  but is not meant to be a public API.

api.d.ts
```diff
@@ -0,0 +1,647 @@
+interface Common {
+    config(new_config: any, new_value: any)
```
takayukioda: Thanks for all the great work!
I appreciate to have this PR so I can skip a lot part of declaring custom type by myself!

Here is some feedback and suggestion that I found while using this type declaration, hope it helps.

I found that config function allows empty argument, which in case sets config if `CLOUDINARY_URL` is been set in environment.
I found configuration parameters in the doc, so added interface for that configuration

```ts
  // https://cloudinary.com/documentation/solution_overview#configuration_parameters
  interface Configuration {
    cloud_name: string
    api_key: string
    api_secret: string
    secure?: boolean
    upload_preset?: string
    cdn_subdomain?: boolean
    private_cdn?: boolean
    cname?: boolean
    secure_distribution?: string
    static_image_support?: boolean
    enhance_image_tag?: boolean
  }

  interface Common {
    config(): Configuration
    config(new_config: keyof Configuration): any
    config(new_config: Configuration): Configuration
    config(new_config: keyof Configuration, new_value: string | boolean): Configuration
```

strausr:
@bd82 Thank you for the contribution. This will be handled in https://github.com/cloudinary/cloudinary_npm/pull/297

PR #273, opened 2019-06-20
Add filename options

strausr:
Add filename options


lib/uploader.js
```diff
@@ -430,10 +430,18 @@ function call_api(action, callback, options, get_params) {
 }
 
 function post(url, post_data, boundary, file, callback, options) {
+  //console.log(if(options.filename) != null {});
```
asisayag2: @strausr, please remove comments.

lib/uploader.js
```diff
@@ -430,10 +430,18 @@ function call_api(action, callback, options, get_params) {
...
   var file_header;
+  let filename;
   let finish_buffer = Buffer.from("--" + boundary + "--", 'ascii');
   if ((file != null) || options.stream) {
-    let filename = options.stream ? "file" : path.basename(file);
+    if (options.filename != null) {
```
asisayag2: @strausr , might not be related to this change, but what happens if the first condition (file!=null || options.stream) is not met?

PR #277, opened 2019-06-25
Exclude version

strausr:
Added exclude version option

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -729,6 +729,7 @@ function url(public_id) {
   var transformation = utils.generate_transformation_string(options);
   var resource_type = utils.option_consume(options, "resource_type", "image");
   var version = utils.option_consume(options, "version");
+  var exclude_version = utils.option_consume(options, "exclude_version");
   var format = utils.option_consume(options, "format");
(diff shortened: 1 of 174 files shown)


asisayag2:
@strausr, why this PR has 173 files changes? 
Also the tests failed.

lib-es5/utils/index.js
```diff
@@ -791,6 +792,9 @@ function url(public_id) {
   if (version != null) {
     version = `v${version}`;
   }
+  if (exclude_version) {
+    version = '';
```
asisayag2: @strausr , please try to simplify the entire version-related logic here... 
In any case, if exclude_version is true, the rest of the logic is irrelevant, so you can start with that check and put the rest in else clause.

test/utils_spec.js
```diff
@@ -1257,6 +1257,10 @@ describe("utils", function () {
   it("should not add version if public_id contains version already", function () {
     test_cloudinary_url("v1234/test", {}, `http://res.cloudinary.com/${cloud_name}/image/upload/v1234/test`, {});
   });
+  it("should not add version if exclude_version:true", function () {
```
asisayag2: @strausr , there are more tests that can be done here. 
For example - exclude_version = false or exclude_version=true and version !=null. 
Please check in other SDKs

PR #280, opened 2019-06-26
Feature/duration to condition video

strausr:
duration to condition video


lib/utils/index.js
```diff
@@ -560,6 +563,7 @@ function generate_transformation_string(options) {
     ["delay", "dl"],
     ["density", "dn"],
     ["duration", "du"],
+    ["initial_duration", "idu"],
```
asisayag2: @strausr, that's not needed here

test/utils_spec.js
```diff
@@ -843,6 +843,22 @@ describe("utils", function () {
           });
           expect(url).to.eql("http://res.cloudinary.com/test123/image/upload/if_w_lt_200,c_fill,h_120,w_80/sample");
         });
+        describe('conditional duration video', function () {
```
asisayag2: @strausr , we need to also test the scenario of:
if: "initialDuration > 30" and assert that it's translated to "idu_gt_30"
Also check "duration" and "initial_duration"

asisayag2:
@strausr , tests are failing. please handle before pushing a new change

PR #290, opened 2019-08-11
Add `force_version` transformation parameter 

strausr:


lib-es5/utils/index.js
```diff
@@ -729,6 +729,10 @@ function url(public_id) {
   var transformation = utils.generate_transformation_string(options);
   var resource_type = utils.option_consume(options, "resource_type", "image");
   var version = utils.option_consume(options, "version");
+  var force_version = utils.option_consume(options, "force_version");
```
asisayag2: @strausr , won't 
var force_version = utils.option_consume(options, "force_version") || true
Do the same?

PR #312, opened 2019-11-11
Structured metadata support

RTLcoil:


RTLcoil:
@strausr The above PR is a work in progress.

Could you explain how to properly sign Upload API requests containing structured metadata? 

If you look at this [skipped test](https://github.com/cloudinary/cloudinary_npm/blob/f9c490e14d3660f4fa5659defdf07ec8d3d839b6/test/structured_metadata_spec.js#L150) and allow it to run you will get an `Invalid Signature` error. How do we properly sign the request?

lib/api.js
```diff
@@ -498,3 +498,38 @@ exports.update_resources_access_mode_by_ids = function update_resources_access_m
 ) {
   return updateResourcesAccessMode(access_mode, "public_ids[]", ids, callback, options);
 };
+
+exports.create_metadata_field = function create_metadata_field(metadata, callback, options = {}) {
```
strausr: Can you please change the callback to match v2 signature

lib/api.js
```diff
@@ -498,3 +498,38 @@ exports.update_resources_access_mode_by_ids = function update_resources_access_m
...
+  return call_api("get", "metadata_fields/" + external_id, {}, callback, options);
+};
+
+exports.update_metadata_field = function update_metadata_field(external_id, callback, options = {}) {
+  const params = only(options, "external_id", "type", "label", "mandatory", "default_value", "validation", "datasource");
+  options.content_type = "json";
+  return call_api("put", "metadata_fields/" + external_id, params, callback, options);
+};
+
+exports.update_metadata_field_datasource = function update_metadata_field_datasource(external_id, callback, options = {}) {
```
strausr: Metadata should be sent as payload and not taken from the options

lib/api.js
```diff
@@ -498,3 +498,38 @@ exports.update_resources_access_mode_by_ids = function update_resources_access_m
...
+
+exports.delete_metadata_field = function delete_metadata_field(external_id, callback, options = {}) {
+  return call_api("delete", "metadata_fields/" + external_id, {}, callback, options);
+};
+
+exports.get_metadata_field = function get_metadata_field(external_id, callback, options = {}) {
+  return call_api("get", "metadata_fields/" + external_id, {}, callback, options);
+};
+
+exports.update_metadata_field = function update_metadata_field(external_id, callback, options = {}) {
```
strausr: Metadata should be sent as payload and not taken from the options

lib/api.js
```diff
@@ -498,3 +498,38 @@ exports.update_resources_access_mode_by_ids = function update_resources_access_m
 ) {
   return updateResourcesAccessMode(access_mode, "public_ids[]", ids, callback, options);
 };
+
```
strausr: Can you please add DOCStrings to the methods?

lib/api.js
```diff
@@ -498,3 +498,38 @@ exports.update_resources_access_mode_by_ids = function update_resources_access_m
...
+  return call_api("put", "metadata_fields/" + external_id, params, callback, options);
+};
+
+exports.update_metadata_field_datasource = function update_metadata_field_datasource(external_id, callback, options = {}) {
+  const params = only(options, "values");
+  options.content_type = "json";
+  return call_api("put", "metadata_fields/" + external_id + "/datasource", params, callback, options);
+};
+
+exports.delete_entries_field_datasource = function delete_entries_field_datasource(external_id, callback, options = {}) {
```
strausr: Method should be called delete_datasource_entries. Should pass in external_ids as payload rather than taken from options

test/structured_metadata_spec.js
```diff
@@ -0,0 +1,160 @@
+const expect = require("expect.js");
```
strausr: Each test should be stand-alone and should not be based on previous tests

test/structured_metadata_spec.js
```diff
@@ -0,0 +1,160 @@
...
+const cloudinary = require("../cloudinary");
+const helper = require("./spechelper");
```
strausr: Need to finish the tests for upload, update, and explicit. This means adding functionality to the uploader. You can take a look at the implementation updateMetadata in Java

test/structured_metadata_spec.js
```diff
@@ -0,0 +1,160 @@
...
+      });
+    });
+  });
+  it("should delete entries in metadata field datasource", function () {
+    const external_ids = [metadata_arr[1].datasource.values[0].external_id];
+    return api.delete_entries_field_datasource(metadata_arr[1].external_id, { external_ids }).then((result) => {
+      expect(result).not.to.be.empty();
+    });
+  });
+  it.skip("should upload image with metadata option", function () {
```
strausr: Why is this test marked as skip? Also, the URL used for this test should not rely on a temporary cloud

lib/api.js
RTLcoil: @strausr I am not sure what you mean. Could you explain this please?

RTLcoil: @strausr a possibly related question:
Should all the new functions be `1` or `2` in `lib/v2/api.js` and `lib/v2/uploader.js`? What does it mean?

test/structured_metadata_spec.js
RTLcoil: This test failed and was thus marked `skip` because the metadata feature flag wasn't activated for our cloud.

It has since been activated and uploading works.
We also fixed what you asked for here.

RTLcoil:
Hey @strausr 

Thanks for the review. We implemented everything you suggested and then some.

There are still a few improvements I would like to get done on the tests, but the main code is ready and I wanted to get this to you ASAP as I know time is a factor.

BTW - I still have one open question to you. What did you mean in 
https://github.com/cloudinary/cloudinary_npm/pull/312#pullrequestreview-318349656 ?

RTLcoil:
Pull request updated.

* Updated private variable names and docstrings to match new reference implementation.
* Added missing function - `restore_metadata_field_datasource`
* Vastly improved tests.

test/spechelper.js
```diff
@@ -11,7 +11,7 @@ const https = require('https');
 const cloudinary = require("../cloudinary");
```
strausr: Please change file name to spec_helper

test/structured_metadata_spec.js
```diff
@@ -0,0 +1,356 @@
...
+  });
+
+  describe("add metadata field", function () {
+    it("should create metadata", function () {
+      const metadataFields = [
+        {
+          external_id: EXTERNAL_ID_CREATE,
+          label: LABEL_STRING_2,
+          type: "string",
+          default_value: "blue",
```
strausr: Should align test values to match PHP. Please remove default_value

test/structured_metadata_spec.js
```diff
@@ -0,0 +1,356 @@
+const expect = require("expect.js");
+const Q = require('q');
+const cloudinary = require("../cloudinary");
+const helper = require("./spechelper");
+
+const TEST_ID = helper.TEST_ID;
+const TEST_TAG = helper.TEST_TAG;
```
strausr: There are many tests missing or incomplete. For example, add test for Test creating an integer, enum, date, set metadata filed. Please align tests to match PHP implementation- cloudinary_php/pull/198

PR #327, opened 2019-12-15
Fix type definition of uploader.upload_stream and uploader.unsigned_upload_stream

hakatashi:
According to the implementation below this should be considered as an error.
https://github.com/cloudinary/cloudinary_npm/blob/b8b7f1a38e793d0a75e48529446e71ddd039933c/lib/uploader.js#L26-L37


strausr:
fixed in https://github.com/cloudinary/cloudinary_npm/pull/331

PR #330, opened 2020-01-15
Fixed isRemoteUrl to correctly detect docx files.

patrick-tolosa:
It was not possible to upload base64 encoded docx(and other) files
An incorrect regex resulted in docx files encoded to base64 to not be considered remoteURLs which prevented a successful upload


lib-es5/uploader.js
```diff
@@ -422,10 +422,14 @@ function parseResult(buffer, res) {
   var result = '';
   try {
     result = JSON.parse(buffer);
+    if (result.error && !result.error.name) {
+      result.error.name = "Error";
+    }
```
asisayag2: @patrick-tolosa how are those related to this PR?

PR #331, opened 2020-01-15
Fix/fixed type def for upload stream

patrick-tolosa:
Fixed upload_stream() type definition

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -838,9 +838,9 @@ declare module 'cloudinary' {
 
             function upload_large(path: string, callback?: ErrorCallBack): Promise<any>;
 
-            function upload_stream(upload_preset: string, options?: UploadApiOptions, callback?: ErrorCallBack): Promise<any>;
-
-            function upload_stream(upload_preset: string, callback?: ErrorCallBack): Promise<any>;
+            function upload_stream(options?: UploadApiOptions, callback?: ErrorCallBack): Promise<any>;
+ 
+            function upload_stream(callback?: ErrorCallBack): Promise<any>;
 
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -838,9 +838,9 @@ declare module 'cloudinary' {
 
             function upload_large(path: string, callback?: ErrorCallBack): Promise<any>;
 
-            function upload_stream(upload_preset: string, options?: UploadApiOptions, callback?: ErrorCallBack): Promise<any>;
-
-            function upload_stream(upload_preset: string, callback?: ErrorCallBack): Promise<any>;
+            function upload_stream(options?: UploadApiOptions, callback?: ErrorCallBack): Promise<any>;
```
asisayag2: @patrick-tolosa , shouldn't we only add '?' to the upload_preset?

patrick-tolosa: It doesn't exist in the function declaration at all, so it doesn't look at like it's optional, it just doesn't exist:

`exports.upload_stream = function upload_stream(callback, options = {})`

asisayag2: When uploading an unsigned upload, it's mandatory. 
@strausr how does that work if the method doesn't get an upload_preset?

patrick-tolosa: @asisayag2  - There's a separate method for unsigned uploads: 

`exports.unsigned_upload_stream = function unsigned_upload_stream(upload_preset, callback, options = {})`

this one does take the upload_preset argument

asisayag2:
@patrick-tolosa , please make sure Travis passes before merging

PR #334, opened 2020-01-20
Refactor/refactor out utils functions

patrick-tolosa:
Refactoring and modernizing the nodeJS SDK   
- Broke up several functions to their own files.
- Created a semantic file structure for utils of the same family (encode, parse)
- Updated old variable declarations from 'var' to 'let'


lib-es5/preloaded_file.js
```diff
@@ -68,7 +71,9 @@ var PreloadedFile = function () {
   }], [{
     key: "split_format",
     value: function split_format(identifier) {
-      var format, last_dot, public_id;
+      var format = void 0,
```
asisayag2: Are we creating a new convention here or init variables with `void 0`?
If we do, maybe worth changing all around and add to lint?

lib-es5/uploader.js
```diff
@@ -458,7 +458,7 @@ function call_api(action, callback, options, get_params) {
   var boundary = utils.random_public_id();
   var errorRaised = false;
   var handle_response = function handle_response(res) {
-    // var buffer;
+    // let buffer;
```
asisayag2: Always good to refactor comments ;-)

lib-es5/utils/defaults.js
```diff
@@ -0,0 +1,78 @@
...
+  ">": 'gt',
+  "<=": 'lte',
+  ">=": 'gte',
+  "&&": 'and',
+  "||": 'or',
+  "*": "mul",
+  "/": "div",
+  "+": "add",
+  "-": "sub"
+};
```
asisayag2: That exists in the consts as well

asisayag2: In general, it looks like consts.js and defaults.js are equal

lib-es5/utils/parsing/toArray.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var isArray = require('lodash/isArray');
```
asisayag2: Can we drop the dependency in lodash in this scenario?

PR #335, opened 2020-01-21
fix: move typescript to devDependencies and update major version

nsine:
At the moment when trying to use cloudinary sdk in typescript project of one of latest version - build fails with a lot of errors. (I assume because of two different major versions of TS)

This PR:
* moves typescript to `devDependencies` where it should actually be
* upgrades typescript to latest major version

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
   "main": "cloudinary.js",
   "dependencies": {
     "lodash": "^4.17.11",
-    "q": "^1.5.1",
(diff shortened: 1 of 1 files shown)


nsine:
@strausr @patrick-tolosa can somebody please have a look why ci fails?
i expect it's not because of this dependency change

patrick-tolosa:
@nsine - Our CI has some internal secrets and thus does not run on external PRs, for security reasons.

PR #336, opened 2020-01-21
fix: type defs for stream upload methods

nsine:
All stream upload methods return stream itself, not promise.
This PR fixes typings for them

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "typescript": "^2.9.2"
   },
   "devDependencies": {
+    "@types/node": "^13.5.0",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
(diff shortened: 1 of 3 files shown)


nsine:
@strausr I think using `ts-stream` is not necessary. Adding `@types/node` would be enough for running tests

nsine:
can smb also look at ci? it fails in both of my PRs: this one and https://github.com/cloudinary/cloudinary_npm/pull/335
what would be the reason?

asisayag2:
@nsine, thanks for the PR. 
We will look into it as soon as possible and this will include making sure the CI is passing.

nsine:
@strausr hm
but i already did it month ago 😅
https://github.com/cloudinary/cloudinary_npm/pull/336/commits/41bc0ef210d3ecc0e6fa54de420d5eac2a546115

PR #337, opened 2020-01-22
refactored test to catch API failures

patrick-tolosa:


test/uploader_spec.js
```diff
@@ -713,24 +713,33 @@ describe("uploader", function () {
...
+    }).catch((err) => {
+      console.log(err);
+      throw new Error('Unsigned_upload failed');
+    });
+
+    expect(uploadResponse.public_id).to.match(/^upload_folder\/[a-z0-9]+$/);
+
+    await cloudinary.v2.api.delete_upload_preset(preset.name).catch((err) => {
+      console.log(err);
+      throw new Error('delete_upload_preset failed');
```
asisayag2: @patrick-tolosa , not sure that needs to fail the test. 
Worst case, we have junk upload presets

test/uploader_spec.js
```diff
@@ -713,24 +713,33 @@ describe("uploader", function () {
...
-      return cloudinary.v2.uploader.unsigned_upload(IMAGE_FILE, preset.name, {
-        tags: UPLOAD_TAGS,
-      });
-    }).then(function ({ public_id }) {
-      expect(public_id).to.match(/^upload_folder\/[a-z0-9]+$/);
-    }).finally(function () {
-      return cloudinary.v2.api.delete_upload_preset(presetName);
+    }).catch((err) => {
+      console.log(err);
+      throw new Error('Create upload preset call failed');
```
asisayag2: Let's set a convention of those error messages - "[function name] failed"
in this case - "create_upload_preset failed"

test/uploader_spec.js
```diff
@@ -713,24 +713,33 @@ describe("uploader", function () {
...
+    });
+
+    let uploadResponse = await cloudinary.v2.uploader.unsigned_upload(IMAGE_FILE, preset.name, {
+      tags: UPLOAD_TAGS,
+    }).catch((err) => {
+      console.log(err);
+      throw new Error('Unsigned_upload failed');
```
asisayag2: "**u**nsigned_upload failed"

test/uploader_spec.js
```diff
@@ -713,24 +713,33 @@ describe("uploader", function () {
...
+    });
+
+    let uploadResponse = await cloudinary.v2.uploader.unsigned_upload(IMAGE_FILE, preset.name, {
+      tags: UPLOAD_TAGS,
+    }).catch((err) => {
+      console.log(err);
+      throw new Error('unsigned_upload failed');
+    });
+
+    expect(uploadResponse.public_id).to.match(/^upload_folder\/[a-z0-9]+$/);
```
asisayag2: @patrick-tolosa , isn't that redundant if the upload response failed?

patrick-tolosa: Well , yes. It's also redundant if the upload_preset failed, however, I prefer to keep `expects` outside of conditions (if, try) to make it clear what the test is testing. (Same reason I try to keep `return` statements outside of conditions)

What do you think?

asisayag2: @patrick-tolosa , I understand the approach and agree it's less readable when the asserts happens inside the `then` clause. 
Since the test will fail anyway when the error is thrown, I'm ok with leaving as-is

PR #354, opened 2020-03-10
Refactor/improve provisioning api tests

patrick-tolosa:
Based on changes in this PR(as it's the same file)
https://github.com/cloudinary/cloudinary_npm/pull/353

This PR improves the tests for provisioning API - GET sub_accounts.
Added tests that use the ids[] and prefix arguments

### DIFF
--- a/test/provisioning/account_spec.js
+++ b/test/provisioning/account_spec.js
@@ -16,6 +16,7 @@ describe('account API - Provisioning', function () {
   let USER_ROLE = 'billing';
   let USER_ID;
   let GROUP_ID;
+  let CLOUD_NAME_PREFIX = 'justaname';
   this.timeout(helper.TIMEOUT_LONG);
 
   before("Setup the required test", async function () {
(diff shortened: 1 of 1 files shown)


test/provisioning/account_spec.js
```diff
@@ -93,6 +94,22 @@ describe('account API - Provisioning', function () {
     });
   });
 
+  it('Get a specific subAccount', async function () {
```
asisayag2: @patrick-tolosa , minor, but consistency is important. 
In the previous test, it's called 'sub account' in this one 'subAccount' and the in the next 'sub-account'

test/provisioning/account_spec.js
```diff
@@ -25,13 +26,13 @@ describe('account API - Provisioning', function () {
     }
 
     // Create a sub account(sub cloud)
-    let res = await cloudinary.provisioning.account.create_sub_account('jutaname' + Date.now(), 'jutaname' + Date.now(), {}, true).catch((err) => {
+    let res = await cloudinary.provisioning.account.create_sub_account(CLOUD_NAME_PREFIX + Date.now(), CLOUD_NAME_PREFIX + Date.now(), {}, true).catch((err) => {
```
asisayag2: @patrick-tolosa , I guess the name has no meaning in this case, but It's better that (if anyone will look at our tests as example of usage) we will set the name and the cloudName to be similar.

test/provisioning/account_spec.js
patrick-tolosa: Looks like a mistake in the original, I'll open a PR to address it

test/provisioning/account_spec.js
patrick-tolosa: What do you mean by that? 

Similar to each other? or should the cloud name be called "cloudName"?

asisayag2: Since you're calling Date.now() twice, you might get different values, hence your cloudName and name will have different values. 
Not very interesting in our case, but if someone will copy paste from here...
Just extract to a variable

patrick-tolosa:
@asisayag2 
@strausr 
Is anything else needed here or can this be merged once tests are passing?

PR #355, opened 2020-03-10
Refactor in a wait period for eager uploads

patrick-tolosa:
The current tests are failing randomally, as the even though we've receieved a response
from the server, the back-end operation is not yet done.
any request to delete the derived asset when it doesn't exist will fail, thus fail the test.

This fix adds a small timeout to the request, we might want to consider a more long-term solution.

However, this usecase is not really common (upload and immediately delete)

### DIFF
--- a/test/api_spec.js
+++ b/test/api_spec.js
@@ -51,6 +51,14 @@ const LABEL_INT_1 = 'metadata_label_1_' + TEST_ID;
 const LABEL_INT_2 = 'metadata_label_2_' + TEST_ID;
 const LABEL_INT_3 = 'metadata_label_3_' + TEST_ID;
 
+function wait(ms = 0) {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve();
+    }, ms);
+  });
+}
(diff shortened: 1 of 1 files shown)


asisayag2:
@patrick-tolosa, what will be a longer term solution? 
I hate sleeps in the code. 

What is the response you get? can we re-try?

patrick-tolosa:
> @patrick-tolosa, what will be a longer term solution?
> I hate sleeps in the code.
> 
> What is the response you get? can we re-try?

@asisayag2  - This is a fix to the asynchronicity of the backend service vs the immediate response we get.

When you upload an asset eagerly, the response you get is synchronous, but the back-end is not done with the request.
An immediate request afterwards to get the asset details will return wrong results.

The alternative is to poll the server in one form or another.

asisayag2:
IMHO, smart polling is more robust than sleeping (which can fail tomorrow when something will load the server)

patrick-tolosa:
No argument, It will be improved shortly in the next iteration on the tests.

PR #356, opened 2020-03-11
Updated call_api in Upload and Admin to send either callback or promise

strausr:


lib-es5/api_client/execute_request.js
```diff
@@ -16,6 +16,7 @@ var extend = utils.extend,
 function execute_request(method, params, auth, api_url, callback) {
   var options = arguments.length > 5 && arguments[5] !== undefined ? arguments[5] : {};
 
+  var didUserPassCB = typeof callback === "function";
```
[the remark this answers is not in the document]
asisayag2: Following our discussion yesterday, let's drop this variable completely

lib-es5/api_client/execute_request.js
```diff
@@ -101,9 +103,10 @@ function execute_request(method, params, auth, api_url, callback) {
             query_params
           }
         };
-        deferred.reject(err_obj.error);
-        if (typeof callback === "function") {
+        if (didUserPassCB) {
           callback(err_obj);
+        } else {
+          deferred.reject(err_obj.error);
```
asisayag2: We agreed to move all of those repeating 'if else' blocks with a 'handleError' method

lib/uploader.js
```diff
@@ -355,11 +355,15 @@ function parseResult(buffer, res) {
   return result;
 }
 
-function call_api(action, callback, options, get_params) {
-  if (typeof callback !== "function") {
-    callback = function () {};
+function handleError(callback, result, deferred, error) {
```
patrick-tolosa: There are a couple of these which are identical.

I think we should either rename to make them distinct or we can try to share them.
Which do you think makes more sense?

strausr: updated

lib-es5/api_client/execute_request.js
```diff
@@ -79,16 +87,17 @@ function execute_request(method, params, auth, api_url, callback) {
...
           }, result));
-        } else {
-          deferred.resolve(result);
         }
-        if (typeof callback === "function") {
+
+        if (callback) {
           callback(result);
+        } else {
+          deferred.resolve(result);
```
asisayag2: Why not calling `handleRequestError`?

lib-es5/uploader.js
```diff
@@ -463,14 +467,12 @@ function call_api(action, callback, options, get_params) {
...
   var handle_response = function handle_response(res) {
-    // let buffer;
     if (errorRaised) {
 
       // Already reported
     } else if (res.error) {
       errorRaised = true;
-      deferred.reject(res);
-      callback(res);
+      handleUploadAPIError(callback, res, deferred, res);
```
asisayag2: `res.error`?

lib-es5/uploader.js
```diff
@@ -485,26 +487,27 @@ function call_api(action, callback, options, get_params) {
         result = parseResult(buffer, res);
         if (result.error) {
           result.error.http_code = res.statusCode;
-          deferred.reject(result.error);
+          handleUploadAPIError(callback, result, deferred, result.error);
```
asisayag2: @strausr , if the `error` variable is always `result.error`, why not sending only the result object?

lib-es5/uploader.js
```diff
@@ -485,26 +487,27 @@ function call_api(action, callback, options, get_params) {
...
         } else {
           cacheResults(result, options);
-          deferred.resolve(result);
+          if (callback) {
+            callback(result);
+          } else {
+            deferred.resolve(result);
+          }
```
asisayag2: @strausr , why not calling the handleError?

lib-es5/uploader.js
```diff
@@ -485,26 +487,27 @@ function call_api(action, callback, options, get_params) {
...
         }
-        callback(result);
       });
       res.on("error", function (error) {
         errorRaised = true;
-        deferred.reject(error);
-        callback({ error });
+        handleUploadAPIError(callback, { error }, deferred, error);
```
asisayag2: @strausr , why do we need the the {} wrapping?

lib-es5/api_client/execute_request.js
strausr: That's resolving.. it's not the same behavior

lib-es5/uploader.js
strausr: No.. look at how the original error is sent

lib-es5/uploader.js
strausr: in some places, we're just sending the result, like above.

lib-es5/uploader.js
strausr: Resolving. We can make a resolving handler?

lib-es5/uploader.js
strausr: Look at line 498 in the original

PR #359, opened 2020-03-22
Add types for Structured Metadata functions

strausr:


types/index.d.ts
```diff
@@ -774,6 +807,40 @@ declare module 'cloudinary' {
...
+
+            /****************************** Structured Metadata API V2 Methods *************************************/
+
+            function add_metadata_field(field: MetadataField, options?: AdminApiOptions, callback?: ResponseCallback): Promise<MetadataField>;
+
+            function add_metadata_field(field: MetadataField, callback?: ResponseCallback): Promise<MetadataField>;
+
+            function list_metadata_fields(callback?: ResponseCallback, options?: AdminApiOptions): Promise<MetadataFields>;
+
+            function list_metadata_fields(options?: AdminApiOptions): Promise<MetadataFields>;
```
patrick-tolosa: I think there's a difference between the response of MetadataFields and the argument of MetadataFields

The argument MetadataFields can have missing parts, can the response also have missing parts?
For example, if MetadataField.label is optional, when you fetch it from the server, is will that key be missing?

types/index.d.ts
```diff
@@ -555,6 +555,39 @@ declare module 'cloudinary' {
...
+        external_id?: string;
+        type?: string;
+        label?: string;
+        mandatory?: boolean;
+        default_value?: number;
+        validation?: object;
+        datasource?: object;
+    }
+
+    export interface MetadataFields {
```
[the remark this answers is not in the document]
strausr: yup. changing

types/index.d.ts
strausr: Created two separate interfaces. one for the response and one for the options.

types/index.d.ts
```diff
@@ -555,6 +555,43 @@ declare module 'cloudinary' {
...
+        datasource: object;
+
+        [futureKey: string]: any;
+    }
+
+    export interface MetadataFields {
+        metadata_fields: MetadataFieldApiResponse[]
+    }
+
+    export interface datasourceChange {
```
patrick-tolosa: Capital D!

types/index.d.ts
```diff
@@ -555,6 +555,43 @@ declare module 'cloudinary' {
...
+        label: string;
+        mandatory: boolean;
+        default_value: number;
+        validation: object;
+        datasource: object;
+
+        [futureKey: string]: any;
+    }
+
+    export interface MetadataFields {
```
patrick-tolosa: MetadataFieldsApiResponse maybe?

types/index.d.ts
```diff
@@ -80,10 +79,6 @@ declare module 'cloudinary' {
         | "assist_colorblind"
         | "recolor"
         | "tint"
-        | "contrast"
```
asisayag2: @strausr. why were those removed?

types/cloudinary_ts_spec.ts
```diff
@@ -568,6 +568,42 @@ cloudinary.v2.api.usage(function (error, result) {
 // $ExpectType Promise<any>
 cloudinary.v2.api.usage({public_id: 'demo'});
 
+cloudinary.v2.api.add_metadata_field({
```
asisayag2: @strausr , isn't the $ExpectType comment important here?

strausr: No. We can't check the type of the promise bc call_api is not really called and will always return nothing. Its enough to verify the promise with .then() and a sanity check for us that the functions are accepting the proper fields

PR #363, opened 2020-03-26
Verify protocol in `CLOUDINARY_URL` and `CLOUDINARY_ACCOUNT_URL`

RTLcoil:
Validates that `CLOUDINARY_URL` and `CLOUDINARY_ACCOUNT_URL` begin with `cloudinary://` or `account://` respectively.

### DIFF
--- a/test/cloudinary_spec.js
+++ b/test/cloudinary_spec.js
@@ -790,16 +790,77 @@ describe("cloudinary", function () {
     });
   });
   describe("config", function () {
-    var urlBackup = process.env.CLOUDINARY_URL;
-    after(function () {
-      process.env.CLOUDINARY_URL = urlBackup;
-      cloudinary.config(true);
+    let cloudinaryUrlBackup;
+    let accountUrlBackup;
+    before(function () {
+      cloudinaryUrlBackup = process.env.CLOUDINARY_URL;
+      accountUrlBackup = process.env.CLOUDINARY_ACCOUNT_URL;
     });
-    it("should allow nested values in CLOUDINARY_URL", function () {
-      process.env.CLOUDINARY_URL = "cloudinary://key:secret@test123?foo[bar]=value";
-      cloudinary.config(true);
-      const foo = cloudinary.config().foo;
-      expect(foo && foo.bar).to.eql('value');
+    describe("CLOUDINARY_URL", function () {
+      after(function () {
+        process.env.CLOUDINARY_URL = cloudinaryUrlBackup;
+        cloudinary.config(true);
+      });
+      it("should allow nested values in CLOUDINARY_URL", function () {
+        process.env.CLOUDINARY_URL = "cloudinary://key:secret@test123?foo[bar]=value";
+        cloudinary.config(true);
+        const foo = cloudinary.config().foo;
+        expect(foo && foo.bar).to.eql('value');
+      });
+      it("should load a properly formatted CLOUDINARY_URL", function () {
+        process.env.CLOUDINARY_URL = "cloudinary://123456789012345:ALKJdjklLJAjhkKJ45hBK92baj3@test";
+        cloudinary.config(true);
+      });
+      it("should not be sensitive to case in CLOUDINARY_URL's protocol", function () {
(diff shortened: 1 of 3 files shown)


test/cloudinary_spec.js
```diff
@@ -790,16 +790,77 @@ describe("cloudinary", function () {
...
+        cloudinary.config(true);
+        const foo = cloudinary.config().foo;
+        expect(foo && foo.bar).to.eql('value');
+      });
+      it("should load a properly formatted CLOUDINARY_URL", function () {
+        process.env.CLOUDINARY_URL = "cloudinary://123456789012345:ALKJdjklLJAjhkKJ45hBK92baj3@test";
+        cloudinary.config(true);
+      });
+      it("should not be sensitive to case in CLOUDINARY_URL's protocol", function () {
+        process.env.CLOUDINARY_URL = "CLouDiNaRY://123456789012345:ALKJdjklLJAjhkKJ45hBK92baj3@test";
```
patrick-tolosa: This CLOUDINARY_URL is incorrect, right?
The config() won't throw, but in lib/config.js line 53 we check if the protocol equals 'account' or 'cloudinary'.

the protocol here equals to 'CLouDiNaRY', which means the URL won't be parsed.

maybe we should force .toLowerCase() when reading the protocol instead?

RTLcoil: In [line 50](https://github.com/cloudinary/cloudinary_npm/blob/feature/validate-url-env-variables/lib/config.js#L50) we use the Node.js URL module (which is similar to the browser's `URL` interface) to parse the incoming `CLOUDINARY_URL` or `CLOUDINARY_ACCOUNT_URL`. This interface always [presents the protocol as lowercase](https://nodejs.org/api/url.html#url_urlobject_protocol).


Here's the view when debugging. I made sure to explicitly use a `CLouDiNaRY` link and yet the value of `uri.protocol` is all lowercase:

uri_parsing

patrick-tolosa: Fair enough, I'm wondering if that deeper implementation would change, if these tests would fail or pass.

Can we just improve the test a bit to ensure we get the expected config?  (for example that account_id is present)
I think that could make the test itself a bit clearer as well.
In general if possible, adding a meaningful expect statement in the test is always beneficial.

RTLcoil: > Fair enough, I'm wondering if that deeper implementation would change, if these tests would fail or pass.

Do you mean in case the deeper implementation changes to make the protocol case-sensitive in the future? I would argue that for this very reason this test is a good idea. It ensures that the deeper implementation doesn't change or at the very least if it does change, it will raise a flag to make sure it was a conscious choice that was made.

I think it makes sense for the protocol to be and remain case-insensitive. `CLOUDINARY_URL` and `CLOUDINARY_ACCOUNT_URL` are both structured as URIs and [by definition their scheme](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Definition) (the protocol) should be case-insensitive. Even if `CLOUDINARY_URL` and `CLOUDINARY_ACCOUNT_URL` were not designed to follow the strict definition of a URI, a developer using the SDK might reasonably expect them to... so changing this, is a potential pitfall for them.

> Can we just improve the test a bit to ensure we get the expected config? (for example that account_id is present)

But the code doesn't actually ensure that `account_id` is present... for example: if I run this test case:
```javascript
it("should load a properly formatted CLOUDINARY_URL", function () {
  process.env.CLOUDINARY_URL = "cloudinary://";
  cloudinary.config(true);
});
```
It will pass successfully.

Would you like us to extend the scope of this task to also verify additional things about the URI?
This would include:
* The existence and validity of `api_key`, `api_secret`, and `cloud_name` for `CLOUDINARY_URL`.
* The existence and validity of `account_id`, `provisioning_api_key`, and `provisioning_api_secret` for `CLOUDINARY_ACCOUNT_URL`.

patrick-tolosa: I guess I'm referring to how 

```javascript 
it("should load a properly formatted CLOUDINARY_URL", function () {
  process.env.CLOUDINARY_URL = "foobar://";
  cloudinary.config(true);
});```

would also pass. 
We'll handle it as a separate issue, as it's not the scope of this PR.

PR #368, opened 2020-04-11
Add support for 32 char SHA-256 signatures

RTLcoil:


lib/utils/index.js
```diff
@@ -554,6 +554,7 @@ const URL_KEYS = [
   'cloud_name',
   'cname',
   'format',
+  'long_url_signature',
```
patrick-tolosa, line 557: Since this URL represents a Boolean, 
can we rename it to something that reflects that?

Something like use_long_url_signature maybe?

test/cloudinary_spec.js
```diff
@@ -775,6 +775,23 @@ describe("cloudinary", function () {
...
+    result = cloudinary.utils.url("sample.jpg", options);
+    expect(result).to.eql('http://res.cloudinary.com/test123/image/upload/s--2hbrSMPOjj5BJ4xV7SgFbRDevFaQNUFf--/sample.jpg');
+  });
+  it("should generate urls with a 8 character signature when sign_url is true", function () {
+    var options, result;
+    options = {
+      sign_url: true,
+    };
+    result = cloudinary.utils.url("sample.jpg", options);
+    expect(result).to.eql('http://res.cloudinary.com/test123/image/upload/s--v2fTPYTu--/sample.jpg');
```
patrick-tolosa, line 793: can we generate this signature externally in the test, instead of hardcoding the result?
I think it will make it a bit easier to know what's being tested.
For example, it's hard to tell at a glance if s--v2fTPYTu-- is even a valid signature.

lib/utils/index.js
RTLcoil: The variable name was defined in the task description. We did not change it to comply with the spec and with all the other SDKs which use the same variable name.

From the spec:

CleanShot 2020-04-14 at 15 20 13@2x

test/cloudinary_spec.js
```diff
@@ -180,44 +180,44 @@ describe("cloudinary", function () {
...
-        radius: 10
+        radius: 10,
       })).to.eql(`${upload_path}/r_10/test`);
       expect(cloudinary.utils.url("test", {
-        radius: '10'
+        radius: '10',
       })).to.eql(`${upload_path}/r_10/test`);
       expect(cloudinary.utils.url("test", {
         variables: [['$v', 10]],
         radius: '$v',
```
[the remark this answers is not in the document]
RTLcoil: Updated 👍

test/cloudinary_spec.js
RTLcoil: The spec for this task defines that using certain inputs we can always expects a deterministic outcome producing a certain hardcoded string.

CleanShot 2020-04-14 at 15 25 27@2x

This follows the guidelines for testing across SDKs we got from @const-cloudinary and Asi.

And it makes sense, as generating the signature in the test using the same method that the test is testing defeats the purpose of the test.

In pseudo code this would be akin to
```javascript
expect(sign(str)).to.eql(sign(str))
```

PR #370, opened 2020-04-25
Support for creating folders using Admin API

RTLcoil:
Add `create_folder` to Admin API


RTLcoil:
The changes in `test/api_spec.js` may look big, but its just wrapping the existing _folders_ tests with a new `describe` block and adding the new test for `create_folder`

patrick-tolosa:
Please fix conflicts before we merge

PR #371, opened 2020-04-30
Feature/encode sdk version

patrick-tolosa:
- Implemented SDK encoded analytics
- Fixed bug that prevented a falsy value to be accepted as an argument (ensureOptions.js)
- Added analytics: false option to all tests
- Added tests specifically for analytics.
- In case of any sort of error (Could not build string param), the SDK analytics will be a single letter 'E'


test/utils_spec.js
```diff
@@ -33,6 +33,7 @@ describe("utils", function () {
   afterEach(function () {
     cloudinary.config(defaults({
       secure: null,
+      analytics: false,
```
const-cloudinary, line 36: I see too many `analytics: false`, consider disabling it globally for all tests

patrick-tolosa: I wanted to avoid "if in test mode..." inside our source code. 
Since the configuration is created for every test, I think it's useful to keep it explicit (less magic)

const-cloudinary: this less magic approach is not error proof and will require to have maintenance for each new feature we add(Test class), we also need to consider to disable analytics. and when we have more features like analytics, it can easily go out of control.

For this reason, I suggest one place, where we can disable it for all tests. That can be made by forcing all test to use one base class, or moving this to a helper method, like ResetTestsConfig or something and use it in all methods.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * @description Gets the SDK signature by encoding the SDK version and node version
+ * @param {{isResponsive:boolean}} features
+ * @param {'default' | 'x.y.z' | 'x.y' | string} useSDKVersion Default uses package.json version
+ * @param {'default' | 'x.y.z' | 'x.y' | string} useNodeVersion Default uses process.versions.node
+ * @return {string} encodedSDK Suffix
+ */
+function getSDKSuffix(features = {}, useSDKVersion = 'default', useNodeVersion = 'default') {
```
asisayag2, line 25: Can we find a better name than "suffix"?
Maybe versionStamp/versionID/analyticsParam/whatever.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+function removePatchFromSemver(semVerStr) {
+  let parts = semVerStr.split('.');
+
+  return `${parts[0]}.${parts[1]}`;
+}
+
+/**
+ * @description Gets the SDK signature by encoding the SDK version and node version
+ * @param {{isResponsive:boolean}} features
+ * @param {'default' | 'x.y.z' | 'x.y' | string} useSDKVersion Default uses package.json version
```
asisayag2, line 21: Can it be any string? how the entire thing will behave?

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+function removePatchFromSemver(semVerStr) {
+  let parts = semVerStr.split('.');
+
+  return `${parts[0]}.${parts[1]}`;
+}
+
+/**
+ * @description Gets the SDK signature by encoding the SDK version and node version
+ * @param {{isResponsive:boolean}} features
```
asisayag2, line 20: Should be 'responsive' and not 'isResposnive'. 
How can anyone get the list of available features?

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {'default' | 'x.y.z' | 'x.y' | string} useSDKVersion Default uses package.json version
+ * @param {'default' | 'x.y.z' | 'x.y' | string} useNodeVersion Default uses process.versions.node
+ * @return {string} encodedSDK Suffix
+ */
+function getSDKSuffix(features = {}, useSDKVersion = 'default', useNodeVersion = 'default') {
+  try {
+    // allow to pass a custom SDKVersion
+    let pkgJSONFile = fs.readFileSync(path.join(__dirname, '../../../../package.json'), 'utf-8');
```
asisayag2, line 28: no process.env.{something} to get this?

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+
+    let sdkVersion = useSDKVersion === 'default' ? JSON.parse(pkgJSONFile).version : useSDKVersion;
```
asisayag2, line 30: if we set undefined as default value, we can do something like:
let sdkVersion  = useSDKversion || JSON.parse(...).version;
Right?
Looks more elegant

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+
+    // allow to pass a custom nodeVersion
+    let nodeVersion = useNodeVersion === 'default' ? process.versions.node : useNodeVersion;
+
+    // SDK version should always be in x.y format
```
asisayag2, line 35: x.y.z

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+    let twoPartNodeVersion = removePatchFromSemver(nodeVersion);
+    let encodedSDKVersion = encodeVersion(sdkVersion);
+    let encodedNodeVErsion = encodeVersion(twoPartNodeVersion);
+
+    let featureCode = '0';
+    if (features.isResponsive) {
+      featureCode = 'A';
+    }
```
asisayag2, line 43: Let's store that map somewhere else

lib/utils/encoding/sdkSuffix/revereseVersion.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+function reverseVersion(versionString) {
+  let [major, minor, patch] = versionString.split('.');
+  let reversed = '';
+
+  if (patch) {
+    reversed = `${patch.padStart(2, '0')}${minor.padStart(2, '0')}${major.padStart(2, '0')}`;
+  } else {
+    reversed = `${minor.padStart(2, '0')}${major.padStart(2, '0')}`;
+  }
```
asisayag2, line 15: May i suggest:
reversed = `${minor}${major}`
if (patch) reversed = `${patch}` + reversed

Just to avoid the minor/major duplicate code

lib/utils/ensureOption.js
```diff
@@ -7,10 +7,18 @@
  */
 function defaults(defaultOptions) {
```
asisayag2, line 8: What's that?

test/utils_spec.js
asisayag2: Agree with Costa that we need some global way of disabling this feature for tests, and enable it only when relevant.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: There's currently only a single param in Node (responsive).
There's no way to get the complete set.

I decided to use isResponsive because it's a feature check, and it's a boolean.
This is passed as an argument into this function. 

I really want to avoid passing opaque 'configs' or 'options' around functions, or to use global variables (such as config).

lib/utils/encoding/sdkSuffix/encodeVersion.js
```diff
@@ -0,0 +1,43 @@
+let base64Map = require('./base64Map');
+let reverseVersion = require('./revereseVersion');
+
+/**
+ * @description Encodes a semVer-like version string according to SDK logic.
```
const-cloudinary, line 5: `according to SDK logic.` what does it mean?

lib/utils/encoding/sdkSuffix/encodeVersion.js
```diff
@@ -0,0 +1,43 @@
...
+  let paddedReversed = reversed.padStart(paddedStringLength, '0');
+  // Cast to number 15.05.01 -> 150,501(number)
+  let num = parseInt(paddedReversed);
+  // Represent as binary, add left padding to 12 or 18 characters. 15.05.01 -> 150,501(number)
+  let paddedBinary = num.toString(2).padStart(paddedStringLength, '0');
+
+
+  // turn every 6 bits into a character using the base64Map
+  // Support x.y.z
+  if (parts === 3) {
```
const-cloudinary, line 28: Why don't you just extract it to an array and then:
```
matches.forEach(function(entry) {
    strResult += ${base64Map[entry]};
}
```

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
+let fs = require('fs');
+let path = require('path');
+let encodeVersion = require('./encodeVersion');
+
+let nodeSDKValue = 'M'; // Constant per SDK
```
const-cloudinary, line 5: `nodeSDKValue` ?  maybe let's give it a more generic name, so the code would be much similar with other SDKs.

Like `SDKCode` or `SDKType` or what is written in spec.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
```diff
@@ -0,0 +1,52 @@
...
+
+
+/**
+ * @description Turns x.y.z into x.y
+ * @param {string} semVerStr
+ */
+function removePatchFromSemver(semVerStr) {
+  let parts = semVerStr.split('.');
```
const-cloudinary, line 13: why not `emVerStr.split('.').slice(0, -1).join('.')` ?

lib/utils/encoding/sdkSuffix/revereseVersion.js
const-cloudinary: why not just:
```
versionString.split('.').reverse().forEach(function(part) {
    reversed += `${part.padStart(2, '0')}`;
}
```
?

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: The function can accept x.y as well as x.y.z, I've updated the docs to reflect it better.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: Can't be any string, but 'x.y.z' is not actually a valid input. 
There's no good way to represent it in JSDoc (or even typescript). 

If you provide an invalid input, the SDK analytics string will result to 'E'

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: Refactored to versionID.

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: I'm not a fan of undefined values, especially as this is an internal function that's not intended for external use. yes it's more elegant but also suffers from JS syntax (0, undefined, false, all resolve to falsy).

lib/utils/encoding/sdkSuffix/getSDKSuffix.js
patrick-tolosa: Well if you think about it, we're a library that someone installs.
Even if there was a process.env.{something}, it would probably take the version of the installing project, and not our package.json.

lib/utils/ensureOption.js
patrick-tolosa: There was a bug in this function that didn't accept falsy options when the defaultValue is true.

So for analytics, the default is true, but we want the ability to turn it off externally.

this line of code 
`let value = options[name] || defaultOptions[name] || defaultValue;	`

would always return true (because of the last defaultValue which is true)

lib-es5/utils/encoding/sdkVersionID/encodeVersion.js
```diff
@@ -0,0 +1,44 @@
...
+
+var base64Map = require('./base64Map');
+var reverseVersion = require('./revereseVersion');
+
+/**
+ * @description Encodes a semVer-like version string
+ * @param {string} semVer Input can be either x.y.z or x.y
+ * @return {string}
+ */
+function encodeVersion(semVer) {
```
asisayag2, line 11: @patrick-tolosa , what will happen if I'll send semVer=123456?

lib-es5/utils/encoding/sdkVersionID/revereseVersion.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+/**
+ * @description A semVer like string, x.y.z or x.y is allowed
+ *              Reverses the version positions, x.y.z turns to z.y.x
```
asisayag2, line 5: @patrick-tolosa , would add to the description that it also pad the segments

lib-es5/utils/encoding/sdkVersionID/revereseVersion.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {string} versionString
```
asisayag2, line 6: Why not semVer instead of versionString? just to hint what is expected

lib/utils/ensureOption.js
asisayag2: Sorry, still don't get it. 
Let's discuss on Sunday

lib-es5/utils/encoding/sdkVersionID/base64Map.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+function _toConsumableArray(arr) { if (Array.isArray(arr)) { for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) { arr2[i] = arr[i]; } return arr2; } else { return Array.from(arr); } }
```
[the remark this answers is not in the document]
patrick-tolosa: This is the ES-5 compiled code, we use [...chars] in the source.

lib-es5/utils/encoding/sdkVersionID/encodeVersion.js
patrick-tolosa: There's a test file with a list of invalid tests and their expected results

    `expect(getSDKVersionID({}, 'abcdefg', 'abcdefg')).to.equal('E');
     expect(getSDKVersionID({}, '123456', '123456')).to.equal('E');
     expect(getSDKVersionID({}, '43.21.200', '43.21.26')).to.equal('E');`

When we don't know what to do with the input, we return the constant 'E'

patrick-tolosa:
Changed the analytics key to _s 
Added support for dynamic cloud names (merged test-cloud feature from Master)

PR #372, opened 2020-05-03
Feature/create temp clouds for tests

patrick-tolosa:
- Add command to run tests that run on a fresh new cloud
- Streamlined travis commands to use a single command
- Skip provisioning API tests if the configuration is missing (instead of failing them)


test/api_spec.js
```diff
@@ -852,7 +851,7 @@ describe("api", function () {
   });
   // For this test to work, "Auto-create folders" should be enabled in the Upload Settings.
   // Replace `it` with  `it.skip` below if you want to disable it.
```
strausr, line 853: Why are we skipping all of these tests? Should we not make sure that these clouds have all the needed settings?

patrick-tolosa: Some of our tests have big assumptions about the cloud and account we're testing against. 

Since fixing all these tests is out of scope this PR, we'll skip invalid tests and fix them in near-future PRs

test/integration/api/provisioning/account_spec.js
```diff
@@ -53,6 +54,12 @@ describe('account API - Provisioning', function () {
   });
 
   after('Destroy the sub_account and user that was created', async () => {
+    // Skip 'after' in case we don't have account configuration available
+    // This means that the beforeHook also didn't run
+    let config = cloudinary.config(true);
+    if (!(config.provisioning_api_key && config.provisioning_api_secret && config.account_id)) {
+      return;
+    }
     let delRes = await cloudinary.provisioning.account.delete_sub_account(CLOUD_ID);
```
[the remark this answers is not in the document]
patrick-tolosa: This is for the provisioning API test, which creates clouds for its own testing, (as part of the functionality is the ability to create/destroy clouds)

Our usual tests depend on a cloud to exist, for that we have the create/destroy automation in place.

The provisioning API depends on the creation/destruction of clouds itself.

PR #377, opened 2020-05-06
Replace only() with pick()

patrick-tolosa:
Lodash already has a function that does exactly the same thing,
Since we're already using lodash, it makes sense to use an existing tool.

Also, the name pick() is better than only() for what the function does

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -1,12 +1,12 @@
 const utils = require("./utils");
 const call_api = require("./api_client/call_api");
 
-const { extend, only } = utils;
+const { extend, pick } = utils;
 
 const TRANSFORMATIONS_URI = "transformations";
 
 function deleteResourcesParams(options, params = {}) {
-  return extend(params, only(options, "keep_original", "invalidate", "next_cursor", "transformations"));
+  return extend(params, pick(options, "keep_original", "invalidate", "next_cursor", "transformations"));
 }
 
 exports.ping = function ping(callback, options = {}) {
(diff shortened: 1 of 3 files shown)


RTLcoil:
I would like to bring up two issues, if I may:

---

### 1. `only()` and `pick()` aren't 100% the same, and this might cause a breaking change

The main difference between the two is that `only()` will omit keys when their value is `undefined` or `null`, while `pick` will include them.

CleanShot 2020-05-06 at 20 47 31@2x

What this means in the context of our SDK is that we may now be inadvertently sending additional parameters to the API and that the path of the requests will also change.

For example, consider this API call:

```javascript
cloudinary.v2.api.transformations({
  named: undefined,
  max_results: null,
});
```

Before the change (using `only()`) this resulted in this API request: 

```
href: "https://api.cloudinary.com/v1_1/rtlstudio/transformations?",
path: "/v1_1/rtlstudio/transformations?",
query: (empty string),
search: "?",
```

And after the change (using `pick()`):

```
href: "https://api.cloudinary.com/v1_1/rtlstudio/transformations?max_results=&named=",
path: "/v1_1/rtlstudio/transformations?max_results=&named=",
query: "max_results=&named=",
search: "?max_results=&named=",
```

Are we sure the API knows how to handle this? If so, are there tests that show it knows it can respond - and will continue to respond - correctly to this?

Are we sure this change in the request payload won't affect clients, or should this be considered a breaking change?

---

### 2. This PR will further increase our dependency on lodash

I know efforts have been made in the past to reduce our dependency on lodash and this PR nudges us in the opposite direction.

Is this part of a decision to continue with lodash made by the JavaScript team and @asisayag2?

---

A few more minor issues:
1. Missed another instance of `only` in: https://github.com/cloudinary/cloudinary_npm/blob/5447ec4656316dc70f9943850062d23c3d7ac23f/lib/uploader.js#L201
2. Need to compile ES5 files as well.
3. Typo in the word `compatibly` in https://github.com/cloudinary/cloudinary_npm/blob/5447ec4656316dc70f9943850062d23c3d7ac23f/lib/utils/index.js#L1343

asisayag2:
@patrick-tolosa , WDYS?

patrick-tolosa:
First I'd like to mention this is an excellent review, thank you.

1. Awesome catch, we have no tests that check this scenario, I'm not entirely sure we want to be so magical about how we handle values (for example, if "null" will suddenly be a valid value in the server, we'll have problems)

If we really want to omit undefined and nulls for backward compatibility, we might want to do it in a more semantic way. I'll update the PR with a suggestion and a proper test to catch undefined configs.
I really think we need an alternative to `only`.

2. I'm not sure that we should be against using battle-tested libraries in server-based SDKs.
Consider the only vs pick scenario - most developers know what lodash.pick means, very few could accurately guess what our in-house `only` does. 
On the client, where bundle size is critical, we might be more strict.

3. I usually compile to ES-5 after the PR is approved as it makes the review easier, but I'll make sure it's included  (Plus fix typos!)

RTLcoil:
1. Thanks. It's always fun to delve into the details of how a battle-tested library like lodash works and understand the nitty gritty of its implementation. 🤓 

2. +1 on making the distinction between server and client side code when considering dependencies.

3. Cool. Got it 👍

PR #379, opened 2020-05-07
Rename utility function "only"

patrick-tolosa:
This is an improvement on the name of the utility function 'only'.

The function does two distinct things,
1. implements 'pick' from lodash
2. omits undefined and null from the result

Finding a good name for a function that does two things is always difficult, I suggest we start by giving it a clearer name, and maybe in the future think about splitting it up

### DIFF
--- a/lib-es5/api.js
+++ b/lib-es5/api.js
@@ -4,7 +4,7 @@ var utils = require("./utils");
 var call_api = require("./api_client/call_api");
 
 var extend = utils.extend,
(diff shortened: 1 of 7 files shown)


asisayag2:
@patrick-tolosa , why tests are failing?

patrick-tolosa:
Tests are failing due to unrelated code and an issue with our test-account.
I'll re-run the tests so they'll pass.

There's a different PR to address that.

PR #380, opened 2020-05-07
Feature/support download backup version api

strausr:


lib-es5/api.js
```diff
@@ -111,20 +111,22 @@ exports.resource = function resource(public_id, callback) {
...
   type = options.type || "upload";
   uri = ["resources", resource_type, type, public_id];
-  return call_api("get", uri, only(options, "exif", "colors", "derived_next_cursor", "faces", "image_metadata", "pages", "phash", "coordinates", "max_results"), callback, options);
+  return call_api("get", uri, only(options, "exif", "colors", "derived_next_cursor", "faces", "image_metadata", "pages", "phash", "coordinates", "max_results", "versions"), callback, options);
 };
 
 exports.restore = function restore(public_ids, callback) {
   var options = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
 
+  options.content_type = 'json';
```
asisayag2, line 120: @strausr , why is that needed?

test/api_spec.js
```diff
@@ -953,6 +986,40 @@ describe("api", function () {
       expect(resource).not.to.be(null);
       expect(resource.bytes).to.eql(3381);
     }));
+    it('should restore a deleted resource by versions', function () {
+      this.timeout(TIMEOUT.LARGE);
+      return Q.all([
+        uploadImage({
```
asisayag2, line 992: @strausr , shouldn't we upload 2 versions for the same publicId for checking this scenario?

strausr: Based on the spec, it's a list that maps public_ids to versions. I don't think that you can map one public_id to many versions. That, being said, I'll check to see if that works. If it does, I'll add a separate test.

lib-es5/api.js
strausr: This is done to accept an array as part of the options. Meaning this is how we know to parse an array.

test/api_spec.js
asisayag2: I meant that for restoring an **old** version, you need to have at least 2 versions. The old and the new

lib-es5/api.js
asisayag2: Not sure I get what you mean

patrick-tolosa: By default, we send all our requests URL-encoded (even post requests).

It seems our servers are unable (bug, or by design) to parse arrays sent this way.

We don't have this problem once we change the payload to JSON.
We have this line of code littered throughout the project (setting the content_type).

We might want to investigate why we need this (it exists in other SDKs as well)

test/api_spec.js
strausr: added a new test to restore an older version

test/api_spec.js
```diff
@@ -321,6 +325,36 @@ describe("api", function () {
...
+    }));
+    after(function () {
+      return cloudinary.v2.api.delete_resources(publicId).then((response) => {
+        expect(response).to.have.property("deleted");
+      });
+    });
+    it("should return the asset details together with all of its backed up versions when versions is true", function () {
+      return cloudinary.v2.api.resource(publicId, { versions: true })
+        .then((resource) => {
+          expect(resource.versions).not.to.be(undefined);
```
patrick-tolosa, line 346: I guess these are the same no? if 'resource.versions' is an array, it's also not undefined?

strausr: yeah, I'll remove the undefined one

PR #382, opened 2020-05-07
Fix incorrect test implementation for list resources

patrick-tolosa:


asisayag2:
@patrick-tolosa , what did we try to solve here?

patrick-tolosa:
The 'list-resources-by-prefix' test was implemented incorrectly:
1. Upload a file with a publicID
2. Fetch by a prefix. (part of the publicID)
3. Ensure the publicID exists in the result.

The problem we had is that the account returned more than the limit for a single page, and the public ID we just uploaded sometimes would not return on the same page.


The test was adjusted to make sure all publicIDs in the result match the prefix, instead of looking for a specific publicID

PR #385, opened 2020-05-12
Add tests for new OCR features

RTLcoil:
- Support `ocr` in data structure returned from upload
- Support `ocr` parameter in Upload API (`upload()` and `explicit()`) and Admin API (`update()`)
- Support `gravity` values `ocr_text[:<engine>]` and `auto:ocr_text`
- Support new `pixelate_region` and `blur_region` effects (with gravity support)

### DIFF
--- a/test/integration/api/uploader/uploader_spec.js
+++ b/test/integration/api/uploader/uploader_spec.js
@@ -552,14 +552,6 @@ describe("uploader", function () {
       expect(result.moderation[0].kind).to.eql("manual");
     });
   });
-  it("should support requesting ocr analysis", function () {
-    return cloudinary.v2.uploader.upload(IMAGE_FILE, {
-      ocr: "adv_ocr",
-      tags: UPLOAD_TAGS
-    }).then(function (result) {
-      expect(result.info.ocr).to.have.key("adv_ocr");
-    });
-  });
   it("should support requesting raw conversion", function () {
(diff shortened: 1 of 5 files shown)


patrick-tolosa:
Please fix conflicts so we can merge

test/integration/api/uploader/uploader_spec.js
```diff
@@ -1038,6 +1030,40 @@ describe("uploader", function () {
...
+      });
+
+      // Call explicit on the uploaded image with ocr parameter
+      const explicitResult = await UPLOADER_V2.explicit(uploadResult.public_id, {
+        ocr: ocrType,
+        "tags": [TEST_TAG],
+        type: "upload"
+      });
+
+      // Ensure result isn't an error
```
patrick-tolosa, line 1062: Does this test just ensures that nothing explodes? 
In theory we could pass any number of arguments other than OCR and this test would still work, right/

RTLcoil: From our experimenting, `explicit()` does not return the details of the ocr, ocrType, etc. (unlike the `upload()` api call which does [return it](https://github.com/cloudinary/cloudinary_npm/blob/e3e71a0a2a38da3f002e87c26a17b08130500513/test/integration/api/uploader/uploader_spec.js#L1043))

So all we can do is make sure the `explicit` call does not return an error.

Is there something else that can be done here?

PR #387, opened 2020-05-13
Enabled skipped tests caused by running the suites on a fresh cloud

patrick-tolosa:
- Add timeouts to ensure files have a chance to update
- Fix an implementation bug on the wait() function
- Move wait() to a utility file
- Create a sample.jpg file when creating a new cloud
- Refactor complex failing tests to improve stability


test/api_spec.js
```diff
@@ -927,7 +929,7 @@ describe("api", function () {
         public_id: 'test_folder1/test_subfolder2/item',
         tags: UPLOAD_TAGS,
       }),
-    ]).then(function (results) {
+    ]).then(wait(3000)).then(function (results) {
```
asisayag2, line 932: Why do we need to wait here? 
If the upload returned, the assets supposed to be there

test/api_spec.js
```diff
@@ -1008,17 +1013,16 @@ describe("api", function () {
       return uploadImage({
         public_id: PUBLIC_ID_BACKUP_1,
         backup: true,
-      }).then(() => uploadImage({
+      }).then(wait(1000)).then(() => uploadImage({
```
asisayag2, line 1016: Not sure I get why running on a fresh cloud impacts the time we need to wait until we do something. 
Can you explain?

test/archivespec.js
```diff
@@ -176,7 +176,7 @@ describe("archive", function () {
           sinon.assert.calledWith(write, sinon.match(helper.uploadParamMatcher("target_format", "zip")));
         });
       });
-      it.skip('should create archive with "zip" format and include multiple resource types', function () {
+      it('should create archive with "zip" format and include multiple resource types', function () {
```
asisayag2, line 179: This test still fails in all envs

test/uploader_spec.js
```diff
@@ -560,7 +560,7 @@ describe("uploader", function () {
       expect(result.moderation[0].kind).to.eql("manual");
     });
   });
-  it.skip("should support requesting ocr analysis", function () {
+  it("should support requesting ocr analysis", function () {
```
asisayag2, line 563: Don't see any change here, how come it suddenly working?

test/api_spec.js
patrick-tolosa: We've encountered again and again that it's not the case.

There seems to be some very small, but non-zero delay from the upload success to the assets returning from an API call. 

We didn't encounter these issues as the account already had these assets in place

test/archivespec.js
patrick-tolosa: Worked locally, I've noticed a few more failed on travis, will review.

test/uploader_spec.js
patrick-tolosa: This is due to an account configuration, the account needed an OCR package.

asisayag2: Got it

test/api_spec.js
patrick-tolosa: No definitive answer I'm afraid.

It's either these tests were never 100% stable and this timeout increases their stability,
or maybe the fresh cloud still runs some background things that gives a slight delay to assets being uploaded.

asisayag2: Don't like voodoos.
I suggest checking with someone from the BE to see if your assumption makes sense

patrick-tolosa: This was confirmed by the backend team, while they don't know for a fact what can cause the increased latency for a new cloud, we definitely feel it.

test/api_spec.js
asisayag2: Didn't we say will create a retry mechanism? or you want to keep for another task? if you do, please create a ticket.

patrick-tolosa: Yes it'll be another ticket to resolve these, we have a lot of them all over the code.

PR #388, opened 2020-05-18
updated tests to run linter

strausr:


.eslintrc.js
```diff
@@ -1,102 +1,99 @@
...
-    "vars-on-top": "off",
-    "wrap-iife": "off"
-  }
-}
+    "env": {
+        "browser": true,
+        "commonjs": true,
+        "es6": true
+    },
+    "extends": "airbnb-base",
```
patrick-tolosa, line 7: This should be removed, as this imports all the rules from airbnb. We might want to adopt it, but for now lets turn it off (by deleting it or commenting it out)

.eslintrc.js
```diff
@@ -1,102 +1,99 @@
...
+    "globals": {
+        "Atomics": "readonly",
+        "SharedArrayBuffer": "readonly"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018
+    },
+    "rules": {
+        "arrow-body-style": "off",
```
patrick-tolosa, line 16: I think we should delete all these 'off' rules, lets add later what we want

package.json
```diff
@@ -55,7 +55,7 @@
     "test": "tools/scripts/test.sh",
     "test-with-temp-cloud": "tools/scripts/tests-with-temp-cloud.sh",
     "dtslint": "tools/scripts/ditslint.sh",
-    "lint": "eslint .",
+    "lint": "tools/scripts/lint.sh",
```
patrick-tolosa, line 58: I added lint.sh to prevent running eslint on node 6(which is not supported, and fails in Travis)

tools/scripts/ditslint.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
 dtslint --expectOnly --localTs node_modules/typescript/lib types
+npm run lint
```
patrick-tolosa, line 3: I'm not sure we want to run "npm run lint" after every single command.

I think we should keep it in "npm run test" and "npm run compile/build", All other commands are used within these two. What do you think?

PR #391, opened 2020-05-24
Add support for cinemagraph_analysis parameter in upload, explicit, and resource

RTLcoil:


lib/utils/index.js
```diff
@@ -508,6 +509,9 @@ function updateable_resource_params(options, params = {}) {
   if (options.categorization != null) {
     params.categorization = options.categorization;
   }
+  if (options.cinemagraph_analysis != null) {
+    params.cinemagraph_analysis = options.cinemagraph_analysis;
```
patrick-tolosa, line 513: I think we have a duplicate check here:

Line 283 adds cinemagraph_analysis by force (either true or false, as safebool()), and then we call this function which also does a check if it exists

Can't we just default to send it as false if we don't have it in the options? (to avoid these checks?)
We only use `updateable_resource_params()` from `api` and from `utils` so it should be safe enough to default to false I think.

RTLcoil: Yes, you are right. I removed this second check.

Good catch 👍

patrick-tolosa: Hey,

The failing tests made me take a closer look, as I was interested in why removing this additional check caused all the tests to fail.

Please adjust the following:

1. Please change tests to be integrational instead of mock-based (send a real server request instead of mocking it)
2. We started refactoring of our tests, please use this test as reference. (async await based, looks a bit cleaner).

https://github.com/cloudinary/cloudinary_npm/pull/387/files#diff-32de39551abc083b47743f93da5f75d3R1014

RTLcoil: @patrick-tolosa we're on it.

RTLcoil: Hey @patrick-tolosa 

Pushed the requested changes.

PR #404, opened 2020-06-15
added download_folder method

strausr:


lib-es5/utils/index.js
```diff
@@ -1116,6 +1116,20 @@ function download_zip_url() {
   }));
 }
 
+/**
+ * Returns a URL that when invokes creates an archive of a folder
```
asisayag2, line 1120: creates and downloads

lib-es5/utils/index.js
```diff
@@ -1116,6 +1116,20 @@ function download_zip_url() {
...
+ */
+function download_folder(folder_path) {
+  var options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};
+
+  options.resource_type = "all";
+  options.prefixes = folder_path;
+  var cloudinary_params = exports.sign_request(exports.archive_params(merge(options, {
```
asisayag2, line 1127: How this works? the signature include the `mode='download'` but the options sent to the api does not

lib/utils/index.js
```diff
@@ -1024,6 +1024,18 @@ function download_zip_url(options = {}) {
   }));
 }
 
+/**
```
asisayag2, line 1027: Didn't we drop the need of `lib-es5`?
@patrick-tolosa

test/archivespec.js
```diff
@@ -187,4 +187,38 @@ describe("archive", function () {
...
+      .then((uploadResult)=> {
+        expect(uploadResult).not.to.be(null);
+      });
+
+    after(function () {
+      return cloudinary.api.delete_resources_by_prefix('folder/')
+        .then((deleteResult)=> {
+          expect(deleteResult).not.to.be(null);
+        });
+    });
```
asisayag2, line 202: Why do we need a setup and teardown if we're just checking URLs?

lib-es5/utils/index.js
strausr: its part of the cloudinary_params sent in the hashToQuery(cloudinary_params);

Its the same concept needed in download_archive_url- https://github.com/cloudinary/cloudinary_npm/blob/b1e23fc7c0905777c9c34bf328405b85a00b3577/lib/utils/index.js#L1009

lib-es5/utils/index.js
strausr: okay

lib/utils/index.js
strausr: not merged yet

test/archivespec.js
strausr: I wanted to test it on a real URL but I guess it doesn't matter. Removed.

lib/utils/index.js
```diff
@@ -1024,6 +1024,18 @@ function download_zip_url(options = {}) {
...
+ * Creates and downloads a URL that when invokes creates an archive of a folder
+ */
+function download_folder(folder_path, options = {}) {
+  options.resource_type = "all";
+  options.prefixes = folder_path;
```
nitzanj, line 1032: I wonder what's the correct approach if someone already has prefixes in options.
This is an array, right? could we add the `folder_path` param to it? what happens then, several folders are downloaded...?

lib-es5/utils/index.js
asisayag2: Still don't understand how this works

lib/utils/index.js
strausr: I understand you correctly.. you mean something like this? download_folder('folder1/', {prefixes: ['folder2/']});

In this case, only folder1 is downloaded. The prefixes option is ignored.

nitzanj: yeah, that's what I meant. Not sure it's really important, just an edge case...

lib-es5/utils/index.js
nitzanj: What do you mean? both the signature and the api call include the `mode=download param`.

The `sign_request` method returns the signed params with the `mode` param included (after signing).

asisayag2: What am I missing? 
There are `options` which are sent to the API and when building the signature, we merge `options` with `mode=download`

asisayag2: After discussion - all params are sent on the query string and the options sent to the `api_url` are for general config and not params to the specific api call

PR #415, opened 2020-06-29
added download backedup asset function

strausr:


lib/utils/index.js
```diff
@@ -972,6 +977,22 @@ function zip_download_url(tag, options = {}) {
   return exports.api_url("download_tag.zip", options) + "?" + hashToQuery(params);
 }
 
+/**
+ * The returned url should allow downloading the backedup asset
+ * @param asset_id
+ * @param version_id
+ * @param options
```
asisayag2, line 984: @strausr , please elaborate more on when you can get the `asset_id` and `version_id` from 
Also, would explain more in the description what is the purpose of this method

strausr: updated

PR #419, opened 2020-07-08
Update nodeJS to use cloudinary-core for analytics

patrick-tolosa:


patrick-tolosa:
@strausr  - They're deleted because the source /lib files were deleted.
These files were testing how the mechanics work, but all the functionality moved to core, which has its own tests.

PR #422, opened 2020-07-19
Added wait to versions test

strausr:
Added retry to should restore different versions of a deleted asset

### DIFF
--- a/test/integration/api/admin/api_spec.js
+++ b/test/integration/api/admin/api_spec.js
@@ -1043,6 +1043,8 @@ describe("api", function () {
 
     it('should restore different versions of a deleted asset', async function () {
       this.timeout(TIMEOUT.LARGE);
+      this.retries(3);
+
       // Upload the same file twice (upload->delete->upload->delete)
 
       // Upload and delete a file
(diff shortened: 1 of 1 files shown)


test/integration/api/admin/api_spec.js
```diff
@@ -1043,6 +1043,8 @@ describe("api", function () {
 
     it('should restore different versions of a deleted asset', async function () {
       this.timeout(TIMEOUT.LARGE);
+      this.retries(3);
```
patrick-tolosa, line 1046: There's already a set of retries for in line 144 in this file,
It seems like the retries are too fast, Maybe we need a waiting timer here instead?

strausr: Why do you think the retries are too fast?

patrick-tolosa: Because we already do 3 retries, and this test fails right?

So if 3 retries don't work, it probably means we try too fast

strausr: alright added wait time

patrick-tolosa: Approved, but please change the PR name :)

PR #440, opened 2020-08-13
fix linter errors

strausr:
### Brief Summary of Changes
Fixed linter errors

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -354,15 +354,14 @@ function encode_context(metadataObj) {
       return `${key}=${escapeMetadataValue(value)}`;
 
       // If array, parse each item individually
-    } else if (isArray(value)) {
+    } if (isArray(value)) {
       let values = value.map((innerVal) => {
         return `\"${escapeMetadataValue(innerVal)}\"`
       }).join(',');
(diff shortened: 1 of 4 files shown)


test/spechelper.js
```diff
@@ -217,9 +217,8 @@ exports.provideMockObjects = function (providedFunction) {
 
     if (result && isFunction(result.then)) {
       return result.then(resolve);
-    } else {
-      return resolve(result);
     }
+    return resolve(result);
```
patrick-tolosa, line 221: what lint rule caused this change?
I'm not a huge fan of that

lib/utils/index.js
```diff
@@ -354,15 +354,14 @@ function encode_context(metadataObj) {
...
+    } if (isArray(value)) {
       let values = value.map((innerVal) => {
         return `\"${escapeMetadataValue(innerVal)}\"`
       }).join(',');
       return `${key}=[${values}]`
       // if unknown, return the value as string
-    } else {
-      return value.toString();
     }
+    return value.toString();
```
patrick-tolosa, line 364: Same comment as below, which rule change caused this?

PR #453, opened 2020-09-24
Add missing public_ids[] param for resources function

KadenceNeuens:
### Brief Summary of Changes

There is a public_ids option in the Admin API docs that is missing functionality in the resources function. Added public_ids to the param for the function.

Section where this is found:
https://cloudinary.com/documentation/admin_api#get_resources

- [ ] Refactoring
- [ ] New feature
- [x] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [x] No

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -22,17 +22,21 @@ exports.resource_types = function resource_types(callback, options = {}) {
 };
 
 exports.resources = function resources(callback, options = {}) {
-  let resource_type, type, uri;
+  let params, resource_type, type, uri;
   resource_type = options.resource_type || "image";
   type = options.type;
   uri = ["resources", resource_type];
+  params = pickOnlyExistingValues(options, "next_cursor", "max_results", "prefix", "tags", "context", "direction", "moderations", "start_at");
   if (type != null) {
     uri.push(type);
   }
   if ((options.start_at != null) && Object.prototype.toString.call(options.start_at) === '[object Date]') {
     options.start_at = options.start_at.toUTCString();
   }
-  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "prefix", "tags", "context", "direction", "moderations", "start_at"), callback, options);
+  if (options.public_ids != null) {
+    params["public_ids[]"] = options.public_ids;
+  }
+  return call_api("get", uri, params, callback, options);
 };
 
 exports.resources_by_tag = function resources_by_tag(tag, callback, options = {}) {


patrick-tolosa:
Hey @KadenceNeuens, Thanks for this PR.

The SDK already has a function called resources_by_ids:

```javascript
cloudinary.v2.api.resources_by_ids([publicID1, publicID2], options, cb)
```

Is there a specific reason you also need this functionality on the api.resources function?

We're taking an internal look at our documentation around this area, but if you have a specific use case that would help decide if we'd like to support this or not.

patrick-tolosa (commented):
See previous comment

KadenceNeuens:
Hey @patrick-tolosa , thanks for your time!

The **api.resources** function lists `public_ids` as one of its options under the Admin API documentation, so I had thought it was not working as intended. I also had initially overlooked the brief example where **api.resources_by_ids** is mentioned, but wasn't sure if they had separate and unique functionality. 

The **api.resources_by_id** function isn't documented, and accepts a different set of options than **api.resources**, so I think it would be beneficial to future users for documentation to be added. Additionally, if this isn't implemented, it would be help to remove `public_ids` from the list of options under **api.resources**, as well as the reference to it in the `start_at` option.

After looking into it and testing some possibilities, I actually was not able to find a specific use case. It could be useful to have `start_at` or `direction` options available alongside `public_ids`, but after testing, it seems like it's not supported. Those could be added to **api.resources_by_ids** if supported in the future anyways.

Regardless, I hope the outcome is a more positive user experience. Thank you again!

patrick-tolosa:
@KadenceNeuens 

There's no doubt we'll need to update the documentation, we're wondering what would be the best approach.
You're right in saying that **api.resources_by_ids**  is not the same as **api.resources**,  the main difference is the support for server-side paging. 

this is intentional as it's unlikely that **api.resources_by_ids** is used to retrieve thousands of assets.
I'm keeping this PR open as a discussion point, we'll update once some decisions are made.

You raised good points, it's very much appreciated!

patrick-tolosa:
@KadenceNeuens 

We decided that this feature will not be introduced into the resources endpoint (since resources_by_ids is already available).

We will, however, update our documentation to reflect that more clearly.
I'd like to thank you again for raising this and bringing it to our attention

patrick-tolosa:
Hey @KadenceNeuens 

Just an update,
We've clarified this issue in the documentation
https://cloudinary.com/documentation/admin_api#get_resources

PublicID[]: Note: Not supported for SDKs. Instead, use the resources_by_ids SDK method.

PR #493, opened 2021-06-05
Allow 'multi' and 'sprite' with urls, add 'download_generated_sprite' and 'download_multi' methods

YomesInc:
### Brief Summary of Changes
User can now pass ‘urls’ as a source of sprite/multi.

- [ ] Refactoring
- [x] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [x] Yes
- [ ] No

### DIFF
--- a/lib-es5/uploader.js
+++ b/lib-es5/uploader.js
@@ -264,36 +264,83 @@ exports.text = function text(content, callback) {
   });
 };
 
+/**
+ * Generate a sprite by merging multiple images into a single large image for reducing network overhead and bypassing
+ * download limitations.
+ *
+ * The process produces 2 files as follows:
+ * - A single image file containing all the images with the specified tag (PNG by default).
+ * - A CSS file that includes the style class names and the location of the individual images in the sprite.
+ *
+ * @param {String|Object} tag     A string specifying a tag that indicates which images to include or an object
+ *                                which includes options and image URLs.
+ * @param {Function}     callback   Callback function
+ * @param {Object}       options  Configuration options. If options are passed as the first parameter, this parameter
+ *                                should be empty
+ *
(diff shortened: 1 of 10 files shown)


patrick-tolosa (changes requested):
Hey,
Please add a detailed description in the PR of what was changed, and why.

This PR is very complex to be reviewed without any context.
I can see that a refactor was done around the `api_download_url` but I'm not sure I understand why.

YomesInc:
Since implemented feature assumes similar changes to multiple methods, we've added `build_multi_and_sprite_params` similar to other SDKs, e.g. [PHP](https://github.com/cloudinary/cloudinary_php/pull/275/files#diff-3bf9408acd8fd90ee8aa13cc531a76dfce2605b8da79876c9753859527dae718R267).

Another part of the PR is adding download_generated_sprite and download_multi. By looking at it, all download methods have similar logic: set mode=download parameter, sign parameters, build a URL. And this logic applies not only to new download_ methods, but to download_archive_url  as well. Instead of having 3 places doing same steps, we've decided to implement `api_download_url`.

As for tests, we've implemented a couple of helpers:
- beASprite
- beASignedDownloadUrl
- beAMulti
This way we get a deeper check than usually done in unit tests and tests are more explicit about expectations. E.g. by `expect(result).to.beASprite()` we can see that we expect a sprite and an extensive set of checks makes sure that we have a sprite rather than a somewhat similar object. 

That's the reasoning for the changes made. Please let us know if you'd like more specific details for any part of the changes.

patrick-tolosa:
Thanks @YomesInc 
Lets fix the conflicts and then we can review this PR.

patrick-tolosa (approved):

PR #494, opened 2021-06-10
Improve the return type of cloudinary.v2.config() in TypeScript

MrLeebo:
### Brief Summary of Changes

The original typescript definition returns `void` for all `cloudinary.v2.config(...)` calls, but the actual code always returns something, either the config or a property. This makes it so you can't actually use the config getter form in TypeScript without resorting to type casts.

I modified the typescript definitions to return the correct types for the various config calls and added/edited tests to assert them.

  - Calls will generally return the ConfigOptions.

```ts
    let config = cloudinary.v2.config();
    config = cloudinary.v2.config(true);
    config = cloudinary.v2.config({ cloud_name: "my-cloud" });
```

  - A string parameter works like a getter, returning the property.

```ts
    cloudinary.v2.config("private_cdn") === true
```

  - Two parameters acts like a setter, but returns the config object.

```ts
    const config = cloudinary.v2.config("private_cdn", true);
```

  - If the function gets two parameters, but the 2nd is undefined, it acts like a getter instead of making the property undefined.

```ts
    cloudinary.v2.config("private_cdn", undefined) === true
```

- [ ] Refactoring
- [ ] New feature
- [x] Bug fix
- [x] Adds more tests

#### Are Tests Included?
- [x] Yes
- [ ] No

### DIFF
--- a/types/cloudinary_ts_spec.ts
+++ b/types/cloudinary_ts_spec.ts
@@ -1,59 +1,108 @@
-import * as cloudinary from 'cloudinary';
+import * as cloudinary from "cloudinary";
 import * as Http from "http";
 
-// $ExpectType void
-cloudinary.v2.config({
-    cloud_name: 'demo',
-});
+// $ExpectType ConfigOptions
+cloudinary.v2.config();
 
-// $ExpectType void
+// $ExpectType ConfigOptions
 cloudinary.v2.config(true);
 
-// $ExpectType void
-cloudinary.v2.config("private_cdn", true);
+// $ExpectType ConfigOptions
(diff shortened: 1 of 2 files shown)


patrick-tolosa:
Thanks @MrLeebo , can you please revert all the style changes done on `cloudinary_ts_spec.ts`?

MrLeebo:
@patrick-tolosa OK. The formatting changes have been reverted.

patrick-tolosa (approved):

patrick-tolosa:
@strausr please take a look at these new types.

strausr (approved):

PR #508, opened 2021-08-11
Add support for create_slideshow Upload API

patrick-tolosa:
### Brief Summary of Changes
Add an API endpoint for create_slideshow

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/lib-es5/uploader.js
+++ b/lib-es5/uploader.js
@@ -219,6 +219,29 @@ exports.create_zip = function create_zip(callback) {
   return exports.create_archive(callback, options, "zip");
 };
 
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = "video";
+  return call_api("create_slideshow", callback, options, function () {
+    // Generate a transformation from the manifest_transformation key, which should be a valid transformation
+    var manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
+
+    // Try to use all the options to generate a transformation (Example: options.width and options.height)
+    var transformation = utils.generate_transformation_string(extend({}, options));
+
+    return [{
+      timestamp: utils.timestamp(),
+      manifest_transformation: manifest_transformation,
+      upload_preset: options.upload_preset,
+      overwrite: options.overwrite,
+      public_id: options.public_id,
+      notification_url: options.notification_url,
+      manifest_json: options.manifest_json,
+      tags: options.tags,
+      transformation: transformation
+    }];
+  });
+};
+
 exports.destroy = function destroy(public_id, callback) {
   var options = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
 
--- a/lib-es5/uploader.js
+++ b/lib-es5/uploader.js
@@ -34,6 +34,7 @@ var Cache = require('./cache');
 var utils = require("./utils");
 var UploadStream = require('./upload_stream');
 var config = require("./config");
+var ensureOption = require('./utils/ensureOption').defaults(config());
 
 var build_upload_params = utils.build_upload_params,
     extend = utils.extend,
@@ -220,13 +221,13 @@ exports.create_zip = function create_zip(callback) {
 };
 
 exports.create_slideshow = function create_slideshow(options, callback) {
-  options.resource_type = "video";
+  options.resource_type = ensureOption(options, "resource_type", "video");
   return call_api("create_slideshow", callback, options, function () {
     // Generate a transformation from the manifest_transformation key, which should be a valid transformation
     var manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
 
     // Try to use all the options to generate a transformation (Example: options.width and options.height)
-    var transformation = utils.generate_transformation_string(extend({}, options));
+    var transformation = utils.generate_transformation_string(extend({}, ensureOption(options, 'transformation', {})));
 
     return [{
       timestamp: utils.timestamp(),
--- a/lib-es5/v2/uploader.js
+++ b/lib-es5/v2/uploader.js
@@ -35,3 +35,4 @@ exports.upload_tag_params = uploader.upload_tag_params;
 exports.upload_url = uploader.upload_url;
 exports.image_upload_tag = uploader.image_upload_tag;
 exports.unsigned_image_upload_tag = uploader.unsigned_image_upload_tag;
+exports.create_slideshow = uploader.create_slideshow;
--- a/lib/uploader.js
+++ b/lib/uploader.js
@@ -14,6 +14,7 @@ const Cache = require('./cache');
 const utils = require("./utils");
 const UploadStream = require('./upload_stream');
 const config = require("./config");
+const ensureOption = require('./utils/ensureOption').defaults(config());
 
 const {
   build_upload_params,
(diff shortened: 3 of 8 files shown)


strausr (approved):

lib-es5/uploader.js
```diff
@@ -219,6 +219,29 @@ exports.create_zip = function create_zip(callback) {
   return exports.create_archive(callback, options, "zip");
 };
 
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = "video";
```
const-cloudinary, line 223: It would be more future-proof if you set it as a default value.
like:
`ensureOption(options, "resource_type", "video");`

lib-es5/uploader.js
```diff
@@ -219,6 +219,29 @@ exports.create_zip = function create_zip(callback) {
...
+  return call_api("create_slideshow", callback, options, function () {
+    // Generate a transformation from the manifest_transformation key, which should be a valid transformation
+    var manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
+
+    // Try to use all the options to generate a transformation (Example: options.width and options.height)
+    var transformation = utils.generate_transformation_string(extend({}, options));
```
const-cloudinary, line 229: Since this is a new API, I would prefer it to behave according to spec, and check transformation only under `transformation` key, and without "best" effort that we do in other (legacy) apis.

lib-es5/v2/uploader.js
```diff
@@ -34,4 +34,5 @@ exports.direct_upload = uploader.direct_upload;
 exports.upload_tag_params = uploader.upload_tag_params;
 exports.upload_url = uploader.upload_url;
 exports.image_upload_tag = uploader.image_upload_tag;
-exports.unsigned_image_upload_tag = uploader.unsigned_image_upload_tag;
\ No newline at end of file
+exports.unsigned_image_upload_tag = uploader.unsigned_image_upload_tag;
+exports.create_slideshow = uploader.create_slideshow;
```
const-cloudinary, line 38: No new line at the end of file in this case pollutes diff, line 37 should not have been affected

lib/uploader.js
```diff
@@ -167,6 +167,34 @@ exports.create_zip = function create_zip(callback, options = {}) {
...
+
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = "video"
+  return call_api("create_slideshow", callback, options, function () {
+    // Generate a transformation from the manifest_transformation key, which should be a valid transformation
+    const manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
+
+    // Try to use all the options to generate a transformation (Example: options.width and options.height)
+    const transformation = utils.generate_transformation_string(extend({}, options));
+
```
const-cloudinary, line 179: why 2 newlines?

lib/uploader.js
```diff
@@ -167,6 +167,34 @@ exports.create_zip = function create_zip(callback, options = {}) {
   return exports.create_archive(callback, options, "zip");
 };
 
+
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = "video"
```
const-cloudinary, line 172: See comments above

lib/v2/uploader.js
```diff
@@ -33,3 +33,4 @@ exports.upload_tag_params = uploader.upload_tag_params;
 exports.upload_url = uploader.upload_url;
 exports.image_upload_tag = uploader.image_upload_tag;
 exports.unsigned_image_upload_tag = uploader.unsigned_image_upload_tag;
+exports.create_slideshow = uploader.create_slideshow;
```
const-cloudinary, line 36: This is an example of a good file with a newline at the end :)

test/integration/api/uploader/slideshow_spec.js
```diff
@@ -0,0 +1,102 @@
...
+        custom_function: {
+          function_type: 'render',
+          source: slideshowManifest
+        }
+      },
+      transformation: {
+        format: 'auto'
+      },
+      // manifest_json: slideshowManifestJson,
+      tags: [TEST_TAG],
```
const-cloudinary, line 92: When I checked it, I took a veeeery long time to create the slideshow. When you delete the asset which is in "processing" state, is it really deleted?

const-cloudinary (changes requested):

lib/uploader.js
patrick-tolosa: For readability? :)

test/integration/api/uploader/slideshow_spec.js
patrick-tolosa: We're not deleting the asset actually (It will be deleted when the cloud is deleted) - this operation is ASYNC - there is no asset to delete (as you said, it takes a very long time)

lib-es5/uploader.js
patrick-tolosa: Fixed!

lib-es5/uploader.js
patrick-tolosa: Agreed, fixed!

lib/uploader.js
patrick-tolosa: For readability :)

test/integration/api/uploader/slideshow_spec.js
patrick-tolosa: You're right, we're not deleting the asset.
the asset is generated async, this test does not clean up after itself in that regard.

lib-es5/uploader.js
patrick-tolosa: Agreed, fixed!

lib-es5/uploader.js
```diff
@@ -219,6 +220,29 @@ exports.create_zip = function create_zip(callback) {
   return exports.create_archive(callback, options, "zip");
 };
 
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = ensureOption(options, "resource_type", "video");
+  return call_api("create_slideshow", callback, options, function () {
+    // Generate a transformation from the manifest_transformation key, which should be a valid transformation
+    var manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
+
+    // Try to use all the options to generate a transformation (Example: options.width and options.height)
```
const-cloudinary, line 229: Outdated doc string

lib/uploader.js
```diff
@@ -167,6 +168,33 @@ exports.create_zip = function create_zip(callback, options = {}) {
...
 };
 
+
+exports.create_slideshow = function create_slideshow(options, callback) {
+  options.resource_type = ensureOption(options, "resource_type", "video");
+  return call_api("create_slideshow", callback, options, function () {
+    // Generate a transformation from the manifest_transformation key, which should be a valid transformation
+    const manifest_transformation = utils.generate_transformation_string(extend({}, options.manifest_transformation));
+
+    // Try to use all the options to generate a transformation (Example: options.width and options.height)
```
const-cloudinary, line 178: Outdated doc string

const-cloudinary (changes requested):

lib-es5/uploader.js
patrick-tolosa: build file

lib/uploader.js
patrick-tolosa: Fixed, merging!

PR #509, opened 2021-08-11
Fix incorrect normalization of e_preview:duration

patrick-tolosa:
### Brief Summary of Changes
Fix incorrect normalization of e_preview:duration
Before this PR, e_preview:duration was incorrect normalized to e_preview:du.

This PR fixes that by preventing an expression with `:` to be normalized

- [ ] Refactoring
- [ ] New feature
- [X] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [ ] No

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -148,7 +148,10 @@ function textStyle(layer) {
  * @return {Object|String} A normalized String of the input value if possible otherwise the value itself
  */
 function normalize_expression(expression) {
-  if (!isString(expression) || expression.length === 0 || expression.match(/^!.+!$/)) {
+  // Exclude(Don't normalize) expressions that are variable strings foo_!width!
+  // Exclude(Don't normalize) expressions that contain a semicolon (e_preview:duration)
+  // Exclude(Don't normalize) if expression is not a string, or is an empty string
+  if (!isString(expression) || expression.length === 0 || expression.match(/^!.+!$/) || expression.includes(':')) {
     return expression;
   }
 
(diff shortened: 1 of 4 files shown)


strausr (approved):

lib-es5/utils/index.js
```diff
@@ -148,7 +148,10 @@ function textStyle(layer) {
  * @return {Object|String} A normalized String of the input value if possible otherwise the value itself
  */
 function normalize_expression(expression) {
-  if (!isString(expression) || expression.length === 0 || expression.match(/^!.+!$/)) {
+  // Exclude(Don't normalize) expressions that are variable strings foo_!width!
+  // Exclude(Don't normalize) expressions that contain a semicolon (e_preview:duration)
```
const-cloudinary, line 152: An expression candidate (can be a complex string) can have a ":", I would not exclude those.
I would use a look behind approach like in the reference implementation.

test/unit/cloudinary_spec.js
```diff
@@ -879,4 +879,17 @@ describe("cloudinary", function () {
     const result = cloudinary.utils.url("sample", options);
     expect(result).to.contain("$aheight_300,$mywidth_100/c_scale,h_3_mul_ih_add_$aheight,w_3_add_$mywidth_mul_3_add_4_div_2_mul_iw_mul_$mywidth");
   });
+
+  it('should support duration in video preview', () => {
```
const-cloudinary, line 883: I would remove this test and add a test where `normalize_expression` is tested

const-cloudinary (changes requested):

lib-es5/utils/index.js
patrick-tolosa: Fixed

test/unit/cloudinary_spec.js
patrick-tolosa: Fixed, Imported all the tests from the Python SDK.

const-cloudinary (approved):

PR #515, opened 2021-09-07
fix: `verifyNotificationSignature` timestamps are in seconds

rayeeroded:
### Brief Summary of Changes
As described here https://cloudinary.com/documentation/notifications#notification_payload Cloudinary sends notification payloads with a `X-Cld-Timestamp` and `X-Cld-Signature` headers

 - The timestamp is in a seconds, but the signature verification assumes the timestamp is in milliseconds 
 - The `valid_for` parameter is also in seconds, and `Date.now()` returns milliseconds, which causes the calculation to be off.

- [x] Bug fix

#### Are Tests Included?
- [x] No

#### Reviewer, Please Note:
To verify reviewer needs to set the `notificaiton_url` upload an asset and verify that the given timestamp in the header is in seconds.

on the notification handler code verify the signature with the following code
```js
await cloudinary.utils.verifyNotificationSignature(
    JSON.stringify(req.body),
    req.headers['x-cld-timestamp'],
    req.headers['x-cld-signature'])
```

You would notice the timestamp check currently fails but works fine on after this fix.

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -1093,7 +1093,7 @@ function webhook_signature(data, timestamp) {
  * Verifies the authenticity of a notification signature
  *
  * @param {string} body JSON of the request's body
- * @param {number} timestamp Unix timestamp. Can be retrieved from the X-Cld-Timestamp header
+ * @param {number} timestamp Unix timestamp in seconds. Can be retrieved from the X-Cld-Timestamp header
  * @param {string} signature Actual signature. Can be retrieved from the X-Cld-Signature header
  * @param {number} [valid_for=7200] The desired time in seconds for considering the request valid
  *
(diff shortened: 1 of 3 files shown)


lib-es5/utils/index.js
```diff
@@ -1103,7 +1103,7 @@ function verifyNotificationSignature(body, timestamp, signature) {
   var valid_for = arguments.length > 3 && arguments[3] !== undefined ? arguments[3] : 7200;
 
   // verify that signature is valid for the given timestamp
-  if (timestamp < Date.now() - valid_for) {
```
rayeeroded, line 1106: Notice that without the fix, this if is always `false` when timestamp is in seconds (which it is)
Also, the jsdoc mention `valid_for` is in seconds and `Date.now()` returns milliseconds

strausr: Thanks for the contribution:)

strausr (approved):

PR #549, opened 2022-05-21
Add colors type to upload response, fix type

karlhorky:
### Brief Summary of Changes

- Add a missing optional `colors` key to the `UploadApiResponse` type
- Fix the type on the `ResourceResponseType`, as per [the docs](https://cloudinary.com/documentation/analysis_on_upload#semantic_data_extraction)

- [ ] Refactoring
- [ ] New feature
- [x] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [x] No

#### Reviewer, Please Note:

If you would prefer the type `string` instead of using [TypeScript template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) (released in TS 4.1) then I can make this change.

cc @patrick-tolosa

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -591,6 +591,7 @@ declare module 'cloudinary' {
         access_control: Array<string>;
         context: object;
         metadata: object;
+        colors?: [`#${string}`, number][];
 
         [futureKey: string]: any;
     }
@@ -668,7 +669,7 @@ declare module 'cloudinary' {
                 image_metadata: object;
                 faces: number[][];
                 quality_analysis: number;
-                colors: string[][];
+                colors: [`#${string}`, number][];
                 derived: Array<string>;
                 moderation: object;
                 phash: string;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -669,7 +669,7 @@ declare module 'cloudinary' {
                 image_metadata: object;
                 faces: number[][];
                 quality_analysis: number;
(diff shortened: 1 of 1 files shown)


patrick-tolosa:
@karlhorky thanks for the PR.

We'd indeed prefer a string to avoid using TS 4.1, some consumers might still be using old TS versions and thus may run into issues with their tooling.

karlhorky:
Ok, done in 7b55ce6ae19758652336bb3e7d19ef90c70c7bb0

types/index.d.ts
```diff
@@ -668,7 +669,7 @@ declare module 'cloudinary' {
                 image_metadata: object;
                 faces: number[][];
                 quality_analysis: number;
-                colors: string[][];
+                colors: [`#${string}`, number][];
```
patrick-tolosa, line 672: @karlhorky Looks like there's another usage here.

karlhorky: Ah sorry, forgot that I updated two lines in this PR. Fixed in f4e57998c0196ca0aa9b13cff8c74a349a73aeeb

karlhorky:
@patrick-tolosa would you take another look at this?

patrick-tolosa:
Yes of course, sorry I seem to have missed the notification!

karlhorky:
No problem, thanks for the merge!

PR #559, opened 2022-09-04
Add dynamic folder feature

strausr:
### Brief Summary of Changes
Add dynamic folder feature

This PR adds support for the new Admin API function resources_by_asset_folder.

In addition, it adds support for the following parameters:

- asset_folder
- unique_display_name
- use_asset_folder_as_public_id_prefix
- to the relevant admin and upload API functions.

Please note that the integration test for the new function was set with conditional flag, since by default this API is not available for all accounts.

### DIFF
--- a/test/spechelper.js
+++ b/test/spechelper.js
@@ -27,7 +27,7 @@ exports.ICON_FILE = "test/.resources/favicon.ico";
 exports.VIDEO_URL = "http://res.cloudinary.com/demo/video/upload/dog.mp4";
 exports.IMAGE_URL = "http://res.cloudinary.com/demo/image/upload/sample";
 
-exports.ADDON_ALL = 'all'; // Test all addons.
+const ADDON_ALL = 'all'; // Test all addons.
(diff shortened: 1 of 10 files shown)


test/spechelper.js
```diff
@@ -299,4 +300,15 @@ exports.shouldTestAddOn = function (addOn) {
...
 
+/**
+ * Should a certain feature be tested?
+ *
+ * @param {string} feature The feature to test.
+ *
+ * @return boolean
+ */
+exports.shouldTestFeature = function(feature){
+  return feature === 'ALL';
```
const-cloudinary, line 311: I would expect something like with testing addons above ^... this function always returns false in the code...

const-cloudinary (changes requested):

test/spechelper.js
strausr: updated

adimiz1 (approved):

const-cloudinary (approved):

PR #567, opened 2022-11-01
Make types more strict

karlhorky:
### Brief Summary of Changes

Make types more strict, to disallow typos on common literal string option values. Without these changes, there is no type safety for correct option values to be used as transformation options.

- [ ] Refactoring
- [ ] New feature
- [x] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [x] No

#### Reviewer, Please Note:

cc @patrick-tolosa

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -6,7 +6,6 @@ declare module 'cloudinary' {
     /****************************** Constants *************************************/
     /****************************** Transformations *******************************/
     type CropMode =
-        string
         | "scale"
         | "fit"
         | "limit"
@@ -21,7 +20,6 @@ declare module 'cloudinary' {
         | "imagga_crop"
         | "imagga_scale";
     type Gravity =
-        string
         | "north_west"
         | "north"
         | "north_east"
@@ -61,7 +59,6 @@ declare module 'cloudinary' {
         | "ocr_text";
     type Angle = number | string | Array<number | string> | "auto_right" | "auto_left" | "ignore" | "vflip" | "hflip";
     type ImageEffect =
-        string
         | "hue"
         | "red"
         | "green"
@@ -166,7 +163,6 @@ declare module 'cloudinary' {
         | "tiff8_lzw"
         | "tiled";
     type VideoFlags =
-        string
         | Array<string>
         | "animated"
         | "awebp"
@@ -181,9 +177,8 @@ declare module 'cloudinary' {
         | "relative"
         | "truncate_ts"
         | "waveform";
-    type ColorSpace = string | "srgb" | "no_cmyk" | "keep_cmyk";
+    type ColorSpace = "srgb" | "no_cmyk" | "keep_cmyk";
     type DeliveryType =
-        string
         | "upload"
         | "private"
         | "authenticated"
@@ -204,9 +199,8 @@ declare module 'cloudinary' {
         | "worldstarhiphop"
         | "dailymotion";
     /****************************** URL *************************************/
-    type ResourceType = string | "image" | "raw" | "video";
+    type ResourceType = "image" | "raw" | "video";
     type ImageFormat =
-        string
         | "gif"
         | "png"
         | "jpg"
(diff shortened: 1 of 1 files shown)


patrick-tolosa:
@karlhorky Thanks for the PR.
The reason these string types exist, even though if they're less accurate, is to accommodate simple future proofing when the new features are added to the API.

At the end, the SDK is a proxy to the transformation API, and we'd like to avoid forcing people to update the SDK just to use a new transformation API feature.

Yes the SDK will include all the future changes, but it's easier for existing users to use the looser types when updates are introduced.

karlhorky:
I was thinking that this was the original reasoning behind this, but it does not help with TypeScript tooling to have `string | <other literals>`

With this change, the `cloudinary` library behaves more like untyped JavaScript. You are losing features like:

- Preventing typos
- Autocomplete

To demonstrate this:

### With `string`

No real autocomplete on possible values - it only knows about `abc` and `def` because they are literals within the same file)

Screenshot 2022-11-02 at 10 54 46

https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDeAoKUCGB+AXFAzsAE4CWAdgOZQA+UA5BgEYDGdN9AJhAGZ0A0KAL4oUzAPZlCmPHESp0GPHTpCUQA

### Using only literal types in the union

Autocomplete works, detection of typos works, etc

Screenshot 2022-11-02 at 10 54 59

https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDeAoKUCGB+AXFA5BgEYDG+UAPgQCYQBm+ANCgL4ookD2AdgM7CY8cRKnQY8+fKxRA

patrick-tolosa:
I honestly believe you're right, but the issue isn't with the quickness of updating this project.

In fact, many people still use old versions of this SDK, and are unable (or unwilling for their own good reasons) to update the SDK.


If we block future proofing, sometime down the line a person would not be able to use a new API feature because his SDK is out-dated, and perhaps that person cannot update his SDK at this time. 
That's the use-case we're aiming to address with future proofing.

This will mean that we'll be forcing them to handle TS errors on their end (with TS-ignore) if they'd like to use new features that don't exist yet in their old version.

karlhorky:
Yeah, this is a common thing in software in general - it becomes out of date and unsupported (also in big, very visible projects such as Node.js, macOS, etc). Deprecating old versions is part of the cycle.

There are many things that could break if using older versions of things, and I think it's not the maintainer's responsibility to support every release of a piece of software indefinitely into the future (that would also be a nightmare - at least with how software is currently built).

So I would suggest that it's ok - and in this case, even desirable - to have an SDK that fits the API like a glove.

patrick-tolosa:
Unfortunately this is where software becomes opinionated, with no one clearly in the right.
It's currently Cloudinary's point of view that supporting older users is more beneficial than having a stricter type definition - I can see why some would disagree with this approach.

However, since this is a matter of opinion I'm closing this PR, if you have a suggestion that improves types while keeping this type of functionality and support in place we'll be more than happy to review it.

Thanks again for wanting to make this more friendly to others!

karlhorky:
> if you have a suggestion that improves types while keeping this type of functionality and support in place we'll be more than happy to review it

You could recommend that any users who want to keep their old version either:

1. stay on version `1.32.0` (if they don't want to upgrade anyway)
2. patch their package with `patch-package` to add back the `string`

Would definitely recommend strongly to not keep these more general types in the codebase, it reduces the effectiveness and point of using TypeScript.

karlhorky:
Hm, just can't see it - I think it's a negative outcome with only a weak potential upside. If there were precedent for this elsewhere with some justification then I'd reconsider, but I don't see any other service / SDK providers that made this decision to neuter the power of literals.

If the library `cloudinary` should decide to ignore all of these warnings and community precedent and still move forward with `string`, `Object` and similar, you might as well remove the literal types altogether 😬

karlhorky:
Just saw [the tweet from @thoughtspile](https://twitter.com/thoughtspile/status/1639240110787919875) and found a way that the literals can still show up in autocomplete:

- https://github.com/cloudinary/cloudinary_npm/pull/595/

PR #577, opened 2022-12-13
Hiding sensitive data on error

dannyv-cloudinary:
### Brief Summary of Changes
Allows users to add `hide_sensitive` to their `cloudinary.config()` in order to ensure API key & secret, and Authorization header info is removed from error responses

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/lib-es5/api_client/execute_request.js
+++ b/lib-es5/api_client/execute_request.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
+
 // eslint-disable-next-line import/order
 var config = require("../config");
 var https = /^http:/.test(config().upload_prefix) ? require('http') : require('https');
@@ -74,12 +76,19 @@ function execute_request(method, params, auth, api_url, callback) {
     request_options.headers['Content-Length'] = Buffer.byteLength(query_params);
   }
   handle_response = function handle_response(res) {
-    const {hide_sensitive = false} = config();
-    const sanitized_request_options = {...request_options};
+    var _config = config(),
+        _config$hide_sensitiv = _config.hide_sensitive,
+        hide_sensitive = _config$hide_sensitiv === undefined ? false : _config$hide_sensitiv;
 
-    if (hide_sensitive === true){
-      if ("auth" in sanitized_request_options) { delete sanitized_request_options.auth; }
-      if ("Authorization" in sanitized_request_options.headers) { delete sanitized_request_options.headers.Authorization; }
+    var sanitized_request_options = _extends({}, request_options);
+
+    if (hide_sensitive === true) {
+      if ("auth" in sanitized_request_options) {
+        delete sanitized_request_options.auth;
+      }
+      if ("Authorization" in sanitized_request_options.headers) {
+        delete sanitized_request_options.headers.Authorization;
+      }
     }
 
     if (includes([200, 400, 401, 403, 404, 409, 420, 500], res.statusCode)) {
@@ -149,7 +158,7 @@ function execute_request(method, params, auth, api_url, callback) {
           request_options: sanitized_request_options,
           query_params,
           hide_sensitive
-      }
+        }
       };
       deferred.reject(err_obj.error);
       if (typeof callback === "function") {
--- a/lib-es5/api_client/execute_request.js
+++ b/lib-es5/api_client/execute_request.js
@@ -80,14 +80,14 @@ function execute_request(method, params, auth, api_url, callback) {
         _config$hide_sensitiv = _config.hide_sensitive,
         hide_sensitive = _config$hide_sensitiv === undefined ? false : _config$hide_sensitiv;
 
-    var sanitized_request_options = _extends({}, request_options);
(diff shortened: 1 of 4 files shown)


victorli-cloudinary:
There are two console.logs in your config.spec.js

test/unit/config.spec.js
```diff
@@ -20,6 +20,32 @@ describe("config", function () {
...
+    expect(config.hide_sensitive).to.eql(true)
+  });
+
+  it("should hide API key and secret upon error when `hide_sensitive` is true", async function () {
+    try {
+      cloudinary.config({hide_sensitive: true});
+      const result = await cloudinary.v2.api.resource("?");
+      expect(result).fail();
+    } catch (err) {
+      console.log(err.request_options);
```
victorli-cloudinary, line 34: There's a console.log here

test/unit/config.spec.js
```diff
@@ -20,6 +20,32 @@ describe("config", function () {
...
+    }
+  });
+
+  it("should hide Authorization header upon error when `hide_sensitive` is true", async function () {
+    try {
+      cloudinary.config({hide_sensitive: true});
+      const result = await cloudinary.v2.api.resource("?", { oauth_token: 'MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI4' });
+      expect(result).fail();
+    } catch (err) {
+      console.log(err.request_options);
```
victorli-cloudinary, line 45: There's a console.log here

lib/api_client/execute_request.js
```diff
@@ -66,6 +66,14 @@ function execute_request(method, params, auth, api_url, callback, options = {})
     request_options.headers['Content-Length'] = Buffer.byteLength(query_params);
   }
   handle_response = function (res) {
+    const {hide_sensitive = false} = config();
+    const sanitized_request_options = {...request_options};
```
cloudinary-pkoniu, line 70: I know the previous line suggests something different, but the convention in js is to use lower camel case for naming the variables, in this case it'd be: `sanitizedRequestOptions`. We can also make it shorter to `sanitizedOptions` because it's clear what those options are from the context where they are used.

lib/api_client/execute_request.js
```diff
@@ -98,8 +106,9 @@ function execute_request(method, params, auth, api_url, callback, options = {})
 
         if (result.error) {
           deferred.reject(Object.assign({
-            request_options,
-            query_params
+            request_options: sanitized_request_options,
+            query_params,
+            hide_sensitive
```
cloudinary-pkoniu, line 111: I wouldn't return the value for `hide_sensitive` in the error. The behaviour we implemented here is by default turned off, so the developer needs to explicitly set it to `true`.

Think about it this way: if we return `hide_sensitive`, why not return the whole config?

test/unit/config.spec.js
```diff
@@ -20,6 +20,30 @@ describe("config", function () {
     cloudinary.config(true);
   });
 
+  it("should support `hide_sensitive` config param", function () {
```
cloudinary-pkoniu, line 23: Forgot to mention it: I'd add another test case that explicitly proves that the default value for that env var is false.

test/unit/config.spec.js
```diff
@@ -20,6 +20,30 @@ describe("config", function () {
...
+      expect(result).fail();
+    } catch (err) {
+      expect(err.request_options).not.to.have.property("auth");
+    }
+  });
+
+  it("should hide Authorization header upon error when `hide_sensitive` is true", async function () {
+    try {
+      cloudinary.config({hide_sensitive: true});
+      const result = await cloudinary.v2.api.resource("?", { oauth_token: 'MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI4' });
```
cloudinary-pkoniu, line 41: Actual value of the oath token is irrelevant, so we can set it to `irrelevant`.

Also, be extra careful about committing secrets to a public repo. They'll be here forever ;)

lib-es5/api_client/execute_request.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
```
cloudinary-pkoniu, line 3: Is that a result of `npm run compile`?

cloudinary-pkoniu (changes requested):
Take a look at the comments: the majority of them are cosmetics but there's one about leaking a token to a public repo.

Other than that, good work.

lib-es5/api_client/execute_request.js
dannyv-cloudinary: It is indeed

cloudinary-pkoniu: Cool, resolving the conversation then ;)

test/unit/config.spec.js
cloudinary-pkoniu: If it was used before, we can extract it as a variable to keep it in a single place and name it accordingly to make sure that everyone knows we're not leaking secrets there.

That being said, if the value of the token is irrelevant for the test, I'd keep `irrelevant`. More explicit.

cloudinary-pkoniu (approved):

PR #584, opened 2023-01-05
fix: isRemoteUrl not working on big files sometimes

matheuswanted:
### Brief Summary of Changes
The idea is to remove the later part of the isRemoteUrl regex since you only need to check only the first characters to know if it's a remote URL or not.

We're having issues with the isRemoteUrl on production in some cases. here is the callstack for the error we're having:
```
RangeError: Maximum call stack size exceeded
 at RegExp.test (<anonymous>)
 at isRemoteUrl (/app/node_modules/cloudinary/lib/utils/isRemoteUrl.js:9:140)
 at /app/node_modules/cloudinary/lib/uploader.js:54:12
 at call_api (/app/node_modules/cloudinary/lib/uploader.js:466:52)
 at Object.upload (/app/node_modules/cloudinary/lib/uploader.js:52:10)
 at Object.upload (/app/node_modules/cloudinary/lib/utils/index.js:1260:21)
```

we're assuming the issue is related to applying the regex to big files when the application is under load with less resources available than usual. We're not able to reproduce the issue locally or with tests using the same files that broke, one piece of evidence to support this is the failing files are uploaded correctly afterward.

- [ ] Refactoring
- [ ] New feature
- [x] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [x] No

### DIFF
--- a/lib-es5/utils/isRemoteUrl.js
+++ b/lib-es5/utils/isRemoteUrl.js
@@ -8,7 +8,7 @@ var isString = require('lodash/isString');
  * @returns {boolean} true if the given url is a remote location or data
  */
 function isRemoteUrl(url) {
-  return isString(url) && /^ftp:|^https?:|^gs:|^s3:|^data:([\w-.]+\/[\w-.]+(\+[\w-.]+)?)?(;[\w-.]+=[\w-.]+)*;base64,([a-zA-Z0-9\/+\n=]+)$/.test(url);
+  return isString(url) && /^ftp:|^https?:|^gs:|^s3:|^data:/.test(url);
 }
 
 module.exports = isRemoteUrl;
--- a/lib/utils/isRemoteUrl.js
+++ b/lib/utils/isRemoteUrl.js
@@ -6,7 +6,7 @@ const isString = require('lodash/isString');
  * @returns {boolean} true if the given url is a remote location or data
  */
 function isRemoteUrl(url) {
-  return isString(url) && /^ftp:|^https?:|^gs:|^s3:|^data:([\w-.]+\/[\w-.]+(\+[\w-.]+)?)?(;[\w-.]+=[\w-.]+)*;base64,([a-zA-Z0-9\/+\n=]+)$/.test(url);
+  return isString(url) && /^ftp:|^https?:|^gs:|^s3:|^data:/.test(url);
 }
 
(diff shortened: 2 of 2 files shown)


hgtruong:
I'm having the same issue and the changes in this PR seems to fix it. It's only failing in our tests and causing some tests to fail but working normally in production. @TalLevAmi or @cloudinary-pkoniu, do you mind taking a look at this? (Sorry if I tagged the wrong person!)

The PDFDocument base64 stream that seem to cause this to error out has the size of `6_536_860` in our tests.

bauerjon:
+1 to get this change or similar merged

From ChatGPT:

"""
The "Maximum call stack size exceeded" error can occur when using regular expressions with large strings because the regex engine has to evaluate the entire string and backtrack repeatedly to find a match. This can cause the stack to fill up quickly if the string is very long, resulting in the error you are seeing.

One way to avoid this error is to use a more efficient regex pattern that does not require backtracking, or to break the input string into smaller chunks and process them separately.

Here are some tips you can try to solve this error:

Optimize your regular expression: If possible, use a more efficient regex pattern that does not require backtracking. For example, if you are searching for a simple string pattern, you can use the indexOf method instead of a regular expression.

Split the input string into smaller chunks: If the input string is too large, you can split it into smaller chunks and process them separately. This will reduce the amount of memory needed and the likelihood of the stack overflow error. You can use the substring method to split the string into chunks of a certain length and process them one at a time.

Increase the stack size limit: If neither of the above solutions work, you can try increasing the stack size limit in your environment. In Node.js, you can do this by using the --stack-size command-line option or the --max-old-space-size option to increase the overall memory limit. However, be aware that this is a temporary solution and may not work for all cases. Increasing the stack size limit can also have performance implications and may not solve the underlying problem.
"""

cloudinary-pkoniu:
Sorry for the huge delay in responding, I'll make sure to improve the response times next time.

Anyway - I'm merging these changes to master and will do a patch release asap.

atdcloud:
@cloudinary-pkoniu this issue ready to be closed? Seems like [54baa55](https://github.com/cloudinary/cloudinary_npm/commit/54baa55951470a79c9c6c4c700925eba1f5cd9de) is already merged.

Issue #590, opened 2023-02-19
Custom Analytics IDs Don't Produce Custom Code


colbyfayock:
## Describe the bug in a sentence or two.

When passing in custom analytics IDs with `sdkCode`, `techVersion`, and `sdkSemver` via `url` options, the values get overridden by the default analytics settings.

I suspect the issue is:

```
let analyticsOptions = getAnalyticsOptions(Object.assign({}, options, sdkVersions));
```

Where `sdkVersion` will always override any values passed in through `options`.

That is, if this is even the right way to set custom analytics :)

Reference: https://github.com/cloudinary/cloudinary_npm/blob/ef6f5ad9eb1c0834c5796b833b4a7247a88cd39a/lib/utils/index.js#L833

Related: https://github.com/cloudinary-devs/gatsby-transformer-cloudinary/issues/229

## Issue Type (Can be multiple)
[ ] Build - Can’t install or import the SDK
[ ] Babel - Babel errors or cross browser issues
[ ] Performance - Performance issues
[ ] Behaviour - Functions aren’t working as expected (Such as generate URL)
[ ] Documentation - Inconsistency between the docs and behaviour
[ ] Incorrect Types - For typescript users who are having problems with our d.ts files
[ ] Other (Specify)

## Steps to reproduce

```
const url = cloudinary.url(publicId, {
    ...
    urlAnalytics: true,
    sdkCode: 'X',
  });
```

Doesn't produce an ID with `X`

## Browsers (if issue relates to UI, else ignore)
[ ] Chrome
[ ] Firefox
[ ] Safari
[ ] Other (Specify)
[x] All

## Versions and Libraries (fill in the version numbers)
Cloudinary_NPM SDK version 1.34.0 (latest)

wissam-khalili:
Hi @colbyfayock ,

Thank you for your input.
I will share this information with the SDK team.

Regards,
Wissam

PR #591, opened 2023-02-21
feat: add option to configure tracked analytics

raae:
### Brief Summary of Changes

Add similar behavior to @cloudinary/url-gen for configuring tracked analytics. Work done on stream with @colbyfayock so he should also be able to answer questions about the code.

- [X] GitHub issue (Add reference - #590)
- [ ] Refactoring
- [x] New feature
- [ ] Bug fix
- [x] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
- View the description in #590
- Added one test but copied the expected tracking code from what it actually gave us, so make sure this is the expected tracking code
- ~Code changes only done in `lib`, not `lib-es5`~ now added
- Putting this as draft as I am not convinced we are done, but would still love a review

CleanShot 2023-02-21 at 21 49 23@2x

### DIFF
--- a/test/unit/sdkAnalytics/imageTagWithAnalytics.spec.js
+++ b/test/unit/sdkAnalytics/imageTagWithAnalytics.spec.js
@@ -73,11 +73,11 @@ describe('Tests for sdk analytics through image tag', function () {
 
     let imgStr = cloudinary.image("hello", {
       format: "png",
-      trackedAnalytics: {
-        sdkCode: "X",
-        sdkSemver: "7.3.0",
-        techVersion: "3.4.7"
-      }
+      urlAnalytics: true,
+      sdkCode: "X",
+      sdkSemver: "7.3.0",
+      techVersion: "3.4.7"
+      
     });
 
     expect(imgStr).to.contain(`src='http://res.cloudinary.com/${TEST_CLOUD_NAME}/image/upload/hello.png?_s=AXAEzGT0`);
--- a/test/unit/sdkAnalytics/imageTagWithAnalytics.spec.js
+++ b/test/unit/sdkAnalytics/imageTagWithAnalytics.spec.js
@@ -49,7 +49,7 @@ describe('Tests for sdk analytics through image tag', function () {
       urlAnalytics: true
     });
 
-    expect(imgStr).to.contain(`src='http://res.cloudinary.com/${TEST_CLOUD_NAME}/image/upload/hello.png?_s=AMAlhAM0`);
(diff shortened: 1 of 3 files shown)


colbyfayock:
Context: the DevX team would like @raae to add custom analytics to the Gatsby plugins that she's maintaining which is not possible

raae:
To match the @cloudinary/url-gen we chose to add the option `trackedAnalytics` to be used as follows:

```js
  let imgStr = cloudinary.image("hello", {
      format: "png",
      trackedAnalytics: {
        sdkCode: "X",
        sdkSemver: "7.3.0",
        techVersion: "3.4.7"
      }
    });
```

Note that it's not needed to explicitly enable analytics by setting `urlAnalytics: true` when using  `trackedAnalytics`.

Let us know if you'd like to see a different implementation. We could also need some pointers on how to make this a complete PR, do we need to add types etc.

colbyfayock:
note to CLD team - i'll create a followup ticket with Docs team to potentially add this though the URL Gen doesnt include Docs for this feature. separate discussion

raae:
With the last commit, it seems to pass all the checks. I'll move this to "ready for review".

raae:
Removed the `trackedAnalytics` configuration object and ensured all options fall back to the value returned by `sdkVersions()`.

Example usage as shown in the added test:
```
let imgStr = cloudinary.image("hello", {
  format: "png",
  urlAnalytics: true,
  sdkCode: "X",
  sdkSemver: "7.3.0",
  techVersion: "3.4.7"
});
```

cloudinary-pkoniu (approved):

test/unit/sdkAnalytics/imageTagWithAnalytics.spec.js
```diff
@@ -65,4 +65,21 @@ describe('Tests for sdk analytics through image tag', function () {
...
+    let imgStr = cloudinary.image("hello", {
+      format: "png",
+      urlAnalytics: true,
+      sdkCode: "X",
+      sdkSemver: "7.3.0",
+      techVersion: "3.4.7"
+      
+    });
+
+    expect(imgStr).to.contain(`src='http://res.cloudinary.com/${TEST_CLOUD_NAME}/image/upload/hello.png?_s=AXAEzGT0`);
```
cloudinary-pkoniu, line 83: Shouldn't we use `_a` instead of `_s`?

raae: I'll go ahead and change it today.

raae: Done!

PR #593, opened 2023-02-28
fix: removing nested nulls from options passed to api, closes #581

### LINKED ISSUE #581: When 'null' is in an array of tags, signature breaks
## Describe the bug in a sentence or two.
When doing
```javascript
tags = ['test', null];
await cloudinary.v2.uploader.replace_tag(tags, public_id, { type: "private" });
```
I get a "Invalid Signature" 401 error.

## Issue Type (Can be multiple)
[ ] Build - Can’t install or import the SDK
[ ] Babel - Babel errors or cross browser issues
[ ] Performance - Performance issues
[X] Behaviour - Functions aren’t working as expected (Such as generate URL)
[ ] Documentation - Inconsistency between the docs and behaviour
[ ] Incorrect Types - For typescript users who are having problems with our d.ts files
[ ] Other (Specify)

## Steps to reproduce
```
import cloudinary from 'cloudinary';
import configuration from 'somewhere';

cloudinary.config({
    ...configuration.secrets.cloudinary,
    secure: true
});

const public_id = 'some public ID here',
tags = ['test', null];
await cloudinary.v2.uploader.replace_tag(tags, public_id, { type: "private" });
```

## Versions and Libraries (fill in the version numbers)
Cloudinary_NPM SDK version 1.33.0
Node - 19.3.0
NPM - 9.2.0

## Further comments
Changing [line 386 in /lib/uploader.js](https://github.com/cloudinary/cloudinary_npm/blob/1255763fd35ce2e1aba70ce757edf6521d63ce93/lib/uploader.js#L386) to
```javascript
params.tag = tag.filter(t => t);
```
Fixes the issue, but I think the real problem lies in how `clear_blank()` (in [/lib/utils/index.js](https://github.com/cloudinary/cloudinary_npm/blob/1255763fd35ce2e1aba70ce757edf6521d63ce93/lib/utils/index.js#L1024)) is implemented, but I didn't have time to dig too much deeper.

cloudinary-pkoniu:
### Brief Summary of Changes
Changes are related only to one of `utils` functions: `clear_blank`. Previously, it didn't filter out nested nulls from options passed to the API.

- [X] GitHub issue (Add reference - #581)
- [ ] Refactoring
- [ ] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
N/A

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -1111,7 +1111,9 @@ function clear_blank(hash) {
         k = _ref29[0],
         v = _ref29[1];
 
-    filtered_hash[k] = v;
+    filtered_hash[k] = v.filter ? v.filter(function (x) {
+      return x;
+    }) : v;
   });
   return filtered_hash;
 }
@@ -1600,6 +1602,18 @@ function hashToQuery(hash) {
  */
 
 function present(value) {
+  // if (typeof value === 'string' || Array.isArray(value)) {
+  //   return value.length > 0;
+  // } else if (typeof value === 'number') {
+  //   return value > 0;
+  // } else if (typeof value === 'object') {
+  //   return true;
+  // } else if (typeof value === 'boolean') {
+  //   return value;
+  // }
+  if (Array.isArray(value)) {
+    return true;
+  }
   return value != null && ("" + value).length > 0;
 }
 
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -1602,18 +1602,6 @@ function hashToQuery(hash) {
  */
 
 function present(value) {
-  // if (typeof value === 'string' || Array.isArray(value)) {
-  //   return value.length > 0;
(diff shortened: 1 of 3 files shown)


lib-es5/utils/index.js
```diff
@@ -1600,6 +1602,18 @@ function hashToQuery(hash) {
  */
 
 function present(value) {
+  // if (typeof value === 'string' || Array.isArray(value)) {
```
pawelphilipczyk-cloudinary, line 1605: Do we need that much of commented code left here?

pawelphilipczyk-cloudinary (approved):

lib-es5/utils/index.js
cloudinary-pkoniu: I left it by accident, updated in the recent commit.

PR #606, opened 2023-05-08
feat: exposing structured metadata rules api

cloudinary-pkoniu:
### Brief Summary of Changes
Exposing structured metadata rules API.

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
Docs: https://cloudinary.com/documentation/conditional_metadata_rules_api

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -616,3 +616,23 @@ exports.reorder_metadata_fields = function reorder_metadata_fields(order_by, dir
   const params = { order_by, direction };
   return call_api("put", ["metadata_fields", "order"], params, callback, options);
 };
+
+exports.list_metadata_rules = function list_metadata_rules(callback, options = {}) {
+  return call_api('get', ['metadata_rules'], {}, callback, options);
+};
+
+exports.add_metadata_rule = function add_metadata_rule(metadata_rule, callback, options = {}) {
+  options.content_type = 'json';
+  const params = pickOnlyExistingValues(metadata_rule, 'metadata_field_id', 'condition', 'result', 'name');
+  return call_api('post', ['metadata_rules'], params, callback, options);
+};
+
+exports.update_metadata_rule = function update_metadata_rule(field_external_id, updated_metadata_rule, callback, options = {}) {
+  options.content_type = 'json';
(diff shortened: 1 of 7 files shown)


lib/api.js
```diff
@@ -616,3 +616,23 @@ exports.reorder_metadata_fields = function reorder_metadata_fields(order_by, dir
   const params = { order_by, direction };
   return call_api("put", ["metadata_fields", "order"], params, callback, options);
 };
+
```
const-cloudinary, line 619: Maybe add doc strings to functions?

types/index.d.ts
```diff
@@ -984,6 +1058,24 @@ declare module 'cloudinary' {
             function restore_metadata_field_datasource(field_external_id: string, entries_external_id: string[], options?: AdminApiOptions, callback?: ResponseCallback): Promise<DatasourceChange>;
 
             function restore_metadata_field_datasource(field_external_id: string, entries_external_id: string[], callback?: ResponseCallback): Promise<DatasourceChange>;
+
+            /****************************** Structured Metadata Rules API V2 Methods *************************************/
+            function add_metadata_field(rule: MetadataRule, options?: AdminApiOptions, callback?: ResponseCallback): Promise<MetadataRuleResponse>;
```
const-cloudinary, line 1063: shouldn't it be `add_metadata_rule`?

const-cloudinary (changes requested):

types/index.d.ts
cloudinary-pkoniu: Yes it should, good catch, I'll fix it.

lib/api.js
cloudinary-pkoniu: Usage of doc strings is not consistent across the sdk, so I haven't added it on purpose.

I'd like to even argue they should be removed completely, because they tend to not be updated when the code changes and may be a cause for confusion if that's the case.

const-cloudinary: Ideally we would like to generate a reference, like in JS SDK:
https://cloudinary.com/documentation/sdks/js/url-gen/index.html
The reason it is not consistent, is that old code did not have it. All the new code is well documented. (at least we try)

PR #620, opened 2023-07-18
Cached search support

cloudinary-pkoniu:
### Brief Summary of Changes
New `to_url()` method added to the `Search` class to support new Cached Search feature.

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [X] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
Unfortunately, lots of changed lines included in the PR are auto-formatting from my IDE. I'll work on improving the config for eslint to avoid these in the future.


lib-es5/v2/search.js
```diff
@@ -111,6 +138,37 @@ var Search = function () {
...
+    key: 'to_url',
+    value: function to_url(ttl, next_cursor) {
+      var options = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
+
+      var apiSecret = 'api_secret' in options ? options.api_secret : config().api_secret;
+      if (!apiSecret) {
+        throw new Error('Must supply api_secret');
+      }
+
+      var signingAlgorithm = options.signature_algorithm || config().signature_algorithm || DEFAULT_SIGNATURE_ALGORITHM;
```
const-cloudinary, line 151: Search URL can be only signed with SHA256

lib-es5/v2/search.js
```diff
@@ -111,6 +138,37 @@ var Search = function () {
...
+
+      var urlTtl = ttl || this._ttl;
+
+      var query = this.to_query();
+
+      var urlCursor = next_cursor;
+      if (query.next_cursor && !next_cursor) {
+        urlCursor = query.next_cursor;
+        delete query.next_cursor;
```
const-cloudinary, line 160: We should always delete next cursor whether it comes from the parameter or inside, since it affects signature.

lib-es5/v2/search.js
```diff
@@ -111,6 +138,37 @@ var Search = function () {
...
+      }
+
+      var data = clear_blank(query);
```
const-cloudinary, line 163: We also sort query by key in order to persist signature between different SDKs, since it is the same query.

lib-es5/v2/search.js
```diff
@@ -111,6 +138,37 @@ var Search = function () {
...
+      var encodedQuery = base64Encode(JSON.stringify(data));
+
+      var urlPrefix = build_distribution_domain(options);
+
+      var signature = compute_hash(`${urlTtl}${encodedQuery}${apiSecret}`, signingAlgorithm, 'hex');
+
+      return urlCursor ? `${urlPrefix}/search/${signature}/${urlTtl}/${encodedQuery}/${urlCursor}` : `${urlPrefix}/search/${signature}/${urlTtl}/${encodedQuery}/`;
```
const-cloudinary, line 170: Please rewrite the code not to duplicate `${urlPrefix}/search/${signature}/${urlTtl}/${encodedQuery}` part

lib/v2/search.js
```diff
@@ -1,5 +1,18 @@
...
+  isEmpty,
+  isNumber,
+  api_sign_request,
+  compact_and_sort,
+  compute_hash,
+  build_distribution_domain,
+  clear_blank
+} = require('../utils');
+const {DEFAULT_SIGNATURE_ALGORITHM} = require("../utils/consts");
+const {sign} = require("crypto");
```
const-cloudinary, line 14: Why do we need this here?

test/integration/api/search/search_spec.js
```diff
@@ -83,7 +85,72 @@ describe("search_api", function () {
...
+        api_key: 'test-key'
+      };
+
+      beforeEach(() => {
+        cloudinary.v2.config(commonUrlOptions);
+      });
+
+      const search = cloudinary.v2.search.expression('resource_type:image').sort_by('public_id', 'asc').max_results(10);
+
+      const defaultSignature = 'b2eb2ae76343207c92c267130111b241c87c0246';
```
const-cloudinary, line 105: Please make sure to use the same values as in the reference implementation, so we get exactly the same results. This will assure that we do not have bugs.

const-cloudinary (changes requested):
@cloudinary-pkoniu, great PR!
Please see a few review comments.

lib/v2/search.js
cloudinary-pkoniu: Typo, the require must've been by the IDE by mistake, good catch.

lib-es5/v2/search.js
cloudinary-pkoniu: Will do.

lib-es5/v2/search.js
cloudinary-pkoniu: Will do. I added a new `utils.sort_object_by_key`

lib-es5/v2/search.js
cloudinary-pkoniu: Got it.

lib-es5/v2/search.js
cloudinary-pkoniu: Changed it to `sha256` always

lib/utils/index.js
```diff
@@ -727,6 +735,24 @@ function patchFetchFormat(options = {}) {
   }
 }
 
+function build_distribution_domain(options) {
```
const-cloudinary, line 738: you can reuse this function in `cloudinary_url()` function, no point to have 2 duplicates of code.

const-cloudinary (approved):

PR #629, opened 2023-09-11
fix: fetch overlay video creates correct transformation

cloudinary-pkoniu:
### Brief Summary of Changes
Fixes creation of transformation string for video overlay with fetch.

- [ ] Refactoring
- [ ] New feature
- [X] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -131,8 +131,8 @@ function textStyle(layer) {
   });
 
   if (layer.hasOwnProperty("font_size" || "font_family") || !isEmpty(keywords)) {
-    if (!layer.font_size) throw `Must supply font_size for text in overlay/underlay`;
-    if (!layer.font_family) throw `Must supply font_family for text in overlay/underlay`;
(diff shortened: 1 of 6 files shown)


lib-es5/utils/index.js
```diff
@@ -222,71 +222,117 @@ function process_if(ifValue) {
...
+      components.push(textOptions);
+    }
+
+    if (!isEmpty(public_id)) {
+      public_id = public_id.replace('/', ':');
       components.push(public_id);
+    }
+
+    if (!isEmpty(text)) {
+      // const pattern = /(\$\([a-zA-Z]\w+\))/g;
```
const-cloudinary, line 305: Where this commented out code is coming from? is it in use? are there tests covering this functionality?

const-cloudinary (changes requested):

lib-es5/utils/index.js
cloudinary-pkoniu: Left it behind by mistake, already removed it.

PR #631, opened 2023-09-20
fix: improved calculation of the signature in url

cloudinary-pkoniu:
### Brief Summary of Changes
- Improved and simplified the `compute_hash` function to reflect the same behaviour as in python's sdk
- Removed empty file
- Improved readability and simplified series of test cases for signed URLs
  - Previously unnecessary abstraction layer was used, instead of explicit `assert` from the standard library. Also, mocha's `describe` is used instead of the unnecessary custom one.

- [ ] Refactoring
- [ ] New feature
- [X] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
This change may seem to be dangerous, but in order to ensure correctness, I'm going to create additional PR to python SDK containing two new specs that I added here: `test/unit/url/sign_url_spec.js:79` and `test/unit/url/sign_url_spec.js:98`.

### DIFF
--- a/lib-es5/utils/index.js
+++ b/lib-es5/utils/index.js
@@ -49,7 +49,6 @@ var encodeDoubleArray = require('./encoding/encodeDoubleArray');
 
 var config = require("../config");
 var generate_token = require("../auth_token");
-var utf8_encode = require('./utf8_encode');
 var crc32 = require('./crc32');
 var ensurePresenceOf = require('./ensurePresenceOf');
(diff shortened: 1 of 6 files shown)


cloudinary-pkoniu:
@const-cloudinary Additional test cases in Python: https://github.com/cloudinary/pycloudinary/pull/359

const-cloudinary (approved):

PR #649, opened 2024-01-22
Custom regions gravity

cloudinary-pkoniu:
### Brief Summary of Changes
Added new `upload` and `explicit` parameter: `regions`.

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/test/unit/cloudinaryUtils/encodeRegions.spec.js
+++ b/test/unit/cloudinaryUtils/encodeRegions.spec.js
@@ -1,0 +1,38 @@
+const assert = require('assert');
+
+const encodeRegions = require('../../../lib/utils/encoding/encodeRegions');
+
+describe('encodeRegions', () => {
+  it('should properly encode an object with single region', () => {
+    const regions = {
+      region1: [[1, 2], [3, 4]]
+    };
+    const encodedRegions = encodeRegions(regions);
+    assert.deepStrictEqual('{region1}1,2,3,4', encodedRegions);
+  });
+
+  it('should properly encode an object with multiple regions', () => {
+    const regions = {
+      region1: [[1, 2], [3, 4]],
+      region2: [[5, 6], [7, 8]]
+    };
+    const encodedRegions = encodeRegions(regions);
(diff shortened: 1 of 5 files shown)


test/unit/cloudinaryUtils/encodeRegions.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  it('should properly encode an object with multiple regions with multiple coordinates', () => {
+    const regions = {
+      region1: [[1, 2], [3, 4], [5, 6]],
+      region2: [[7, 8], [9, 10], [11, 12], [13, 14]]
+    };
+    const encodedRegions = encodeRegions(regions);
+    assert.deepStrictEqual('{region1}1,2,3,4,5,6|{region2}7,8,9,10,11,12,13,14', encodedRegions);
+  });
+
+  it('should throw an error if used incorrectly', () => {
```
pawelphilipczyk-cloudinary, line 32: Lots of good tests! Would be nice to add some tests when regions are not numbers or some other cases of "incorrect".

pawelphilipczyk-cloudinary (approved):
Looks good, I've added a comment to test a bit more for possibly incorrect usage. 

It also seemed to me at first the "encode" name is confusing as there is no URI related encoding happening but I guess that is ok.

const-cloudinary (approved):

strausr (approved):

test/unit/cloudinaryUtils/encodeRegions.spec.js
cloudinary-pkoniu: I wanted to avoid making too much of these validations apart from the most obvious ones - I assume that the backend is doing full validation anyway, I wanted to return the most obviously wrong things asap for faster feedback for the user.
By `regions` I think you meant `RegionCoordinate` right? This will be an easy check for vanilla js, I can add it.

PR #651, opened 2024-02-08
add getting started sample

strausr:
### Brief Summary of Changes
add getting started sample


samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
+import {v2 as cloudinary} from 'cloudinary';
+
+// Update to your credentials
```
jackieros, line 3: In our QS, we recommend exporting the credentials as a .env or directly to export your credentials using an export command.  Do you think we should really put these credentials inline in the same file as the code itself when we're trying to give a repo that customers might assume follows best practices?

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// Ensure that you're registered to auto tagging https://cloudinary.com/documentation/google_auto_tagging_addon
+// Ensure that you're registered to background removal https://cloudinary.com/documentation/cloudinary_ai_background_removal_addon
+//Upload an image with auto tag
+//apply eager transformation
+cloudinary.uploader.upload(
+    "https://res.cloudinary.com/demo/image/upload/v1707306308/cld-docs-hp/walking_woman",
+    {
+        public_id: "walking_woman",
+        categorization: "google_tagging",
```
jackieros, line 18:
```suggestion
        categorization: "google_tagging",
        auto_tagging: 0.8
```

jackieros, line 18: Just need to add the auto-tagging param in addition to categorization to instruct the add-on to actually add tags.  The categorization just tells the add-on to return all the categories it finds.  The `auto-tagging: 0.8` then tells the add-on to find all the 'categories'  that were returned with a confidence level 0.8 or higher and add them as tags.

(I randomly picked 0.8 - we might want to pick 0.9 or 0.95 even for our needs to ensure we get the `overcoat` value, and only the next 2 or 3 additional tags at most - and then the designer should use exactly those values in the image we show.

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+        eager: [{effect: "background_removal"}]
+    })
+    .then((result)=>{
+        console.log(result.info.categorization.google_tagging)}).catch((error)=> {console.log(error)});
+
+// log in to you cloudinary account. Navigate to the newly uploaded `walking_woman` image. Select open on the image
```
jackieros, line 24:
```suggestion
```

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+// and navigate to the Metadata tab. Select Auto tags, select plus on the Overcoat tag to tag your image.
```
jackieros, line 25:
```suggestion
```

jackieros: We shouldn't need these comments anymore after adding the auto-tagging param to the upload command.

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    { public_id: "street" })
+    .then((result)=>{
+        console.log(result)}).catch((error)=> {console.log(error)});
+
+
+//Transformation image
+const url = cloudinary.url("walking_woman", {
+    transformation: [
+        { effect: "background_removal" },
+        { gravity: "auto", crop: "auto", width: 450 },
```
jackieros, line 39: It doesn't make sense to do g_auto and c_auto, when only changing the width, since that will just scale down the whole image and not actually do any smart cropping.  In my discussion today with Keren, we discussed changing it to a square using `ar_1.0` along with the width param.

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+// Ensure that you're registered to auto tagging https://cloudinary.com/documentation/google_auto_tagging_addon
+// Ensure that you're registered to background removal https://cloudinary.com/documentation/cloudinary_ai_background_removal_addon
+//Upload an image with auto tag
+//apply eager transformation
+cloudinary.uploader.upload(
+    "https://res.cloudinary.com/demo/image/upload/v1707306308/cld-docs-hp/walking_woman",
+    {
+        public_id: "walking_woman",
+        categorization: "google_tagging",
+        eager: [{effect: "background_removal"}]
```
jackieros, line 19: Not sure there's any reason to do the background_removal as eager, since it's supported as an on-the-fly transformation?  I think that can be confusing for a new user that we have both when it's not really essential?

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    .then((result)=>{
+        console.log(result)}).catch((error)=> {console.log(error)});
+
+
+//Transformation image
+const url = cloudinary.url("walking_woman", {
+    transformation: [
+        { effect: "background_removal" },
+        { gravity: "auto", crop: "auto", width: 450 },
+        { underlay: "street"},
```
jackieros, line 40: We need to specify a resize transformation (`c_auto,g_auto,ar_1,w_450`) for the underlay also, since you always want to resize the underlay to be the identical size as the delivered base image.

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+        {if: "!Overcoat!_in_tags"},
+        {
+            overlay:
+                { font_family: "Arial", font_size: 25,  letter_spacing: 14, font_weight:"bold", text: "NEW COLLECTION" },
```
jackieros, line 44: Keren mentioned that we want a different text & color for the overlay? 
Also, I think the goal was to use something like g_south_east,x_20,y_20 for the placement of the layer? (just guessing on the x & y values - we should test what's best)

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+            color: "white"
+        },
+        {if: "end"},
+        { format: "auto", quality: "auto"}
+    ]
+});
+
+console.log("this is the transformed url", url);
```
jackieros, line 52:
```suggestion
console.log("this is the full transformation url", url);
```

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
+import {v2 as cloudinary} from 'cloudinary';
+
+// Update to your credentials
+cloudinary.config({
+    cloud_name: '<cloud_name>',
+    api_key: '<api_key>',
+    api_secret: '<api_secret>'
+});
+
+// Ensure that you're registered to auto tagging https://cloudinary.com/documentation/google_auto_tagging_addon
```
jackieros, line 10:
```suggestion
// Make sure that you're registered to the Google Auto-tagging Add-on: https://cloudinary.com/documentation/google_auto_tagging_addon
```

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+// Ensure that you're registered to background removal https://cloudinary.com/documentation/cloudinary_ai_background_removal_addon
```
jackieros, line 11:
```suggestion
// Make sure that you're registered to the Cloudinary AI Background Removal Add-on: https://cloudinary.com/documentation/cloudinary_ai_background_removal_addon
```

jackieros (changes requested):
@strausr @Kerensela - Added in some comments based on the discussion that keren & I had today.

samples/getting_started/src/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    })
+    .then((result)=>{
+        console.log(result.info.categorization.google_tagging)}).catch((error)=> {console.log(error)});
+
+// log in to you cloudinary account. Navigate to the newly uploaded `walking_woman` image. Select open on the image
+// and navigate to the Metadata tab. Select Auto tags, select plus on the Overcoat tag to tag your image.
+
+//Upload an image for you underlay
+cloudinary.uploader.upload(
+    "https://res.cloudinary.com/demo/image/upload/v17073124415555/cld-docs-hp/street.jpg",
```
jackieros, line 29: @Kerensela - I just noticed that this 'original' file is only 450x300px - It's not a best practice to upload small resolution files as originals.  Can we check where this image was taken from and upload a bigger original?

Issue #650, opened 2024-01-31
Uploading file as buffer does not works in production.


nirvikpurkait:
## Describe the bug in a sentence or two.
When uploading a file as an `byteArrayBuffer` in Next.Js application , if the server is in `development`, the file gets uploaded, but if the server is in `production`, it throws error. _(Error screenshot is provided below)_

## Issue Type (Can be multiple)
- Production (works in development, but not in production)

## Steps to reproduce
- Go to this [link](https://codesandbox.io/p/devbox/cloudinary-jtjjns).
- Enter `CLOUDINARY_API_COULD_NAME`, `CLOUDINARY_API_KEY`, `CLOUDINARY_API_SECRET` in a `.env` file.
- Run `npm run dev` , it works (file uploads).
- Build the application, run `npm run build`, then launch application, run `npm start`.
- Try to to upload a file, (throws error).

## Error screenshots
![Screenshot 2024-02-01 003806](https://github.com/cloudinary/cloudinary_npm/assets/93871471/de591e68-30f1-4a7b-8262-a2acbf48dc45)

## Versions and Libraries (fill in the version numbers)
Cloudinary_NPM SDK version
Node - `v20.9.0` (codesandbox), `v20.9.0` (system)
NPM - `9.8.1`(codesandbox), `10.2.4` (system)

## Config Files (Please paste the following files if possible)
``` json
{
  "name": "nextjs",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start",
    "lint": "next lint"
  },
  "dependencies": {
    "cloudinary": "^2.0.0",
    "next": "14.1.0",
    "react": "^18",
    "react-dom": "^18"
  },
  "devDependencies": {
    "@types/node": "^20",
    "@types/react": "^18",
    "@types/react-dom": "^18",
    "autoprefixer": "^10.4.17",
    "eslint": "^8",
    "eslint-config-next": "14.1.0",
    "postcss": "^8",
    "tailwindcss": "^3.3.0",
    "typescript": "^5"
  }
}

```

## Repository
[codesandbox](https://codesandbox.io/p/devbox/cloudinary-jtjjns)

skalahasti-cloudinary:
Hi @nirvikpurkait ,

Thanks for contacting Cloudinary. 

[...]

PR #654, opened 2024-03-05
fix: file and field encoding fixed for next production build

cloudinary-pkoniu:
### Brief Summary of Changes
In next@14, production build stripped the strings from all `\r` characters which breaks manual (low-level) encoding of multipart file upload when using streams.

- [X] GitHub issue (Add reference - #650)
- [ ] Refactoring
- [ ] New feature
- [X] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [X] No


Eianaftyio:
.patch

PR #656, opened 2024-03-25
feat: added support for new api in beta - analyze api

cloudinary-pkoniu:
### Brief Summary of Changes
Added support for new Analyze API (beta):
- new namespace added: `analysis`
  - one method available: `analyze_uri`
  - calls cloudinary api `v2`
  - should return immediately if used incorrectly (basic arguments check for vanilla js)

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [X] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

#### Reviewer, Please Note:
Please pay double attention to type descriptions available in *.d.ts files.

### DIFF
--- a/lib/analysis/index.js
+++ b/lib/analysis/index.js
@@ -1,17 +1,30 @@
 const utils = require("../utils");
 const {call_analysis_api} = require('../api_client/call_analysis_api');
 
-function analyze(uri, analysis_type, options = {}, callback) {
-  let params = {
+function analyze_uri(uri, analysis_options, options = {}, callback) {
+  const analysisType = analysis_options.analysis_type;
+
+  const params = {
     uri,
-    analysis_type,
-    parameters: options.analyze_parameters
-  };
+    analysis_type: analysisType
+  }
+
+  if (analysisType === 'custom') {
+    if (!('model_name' in analysis_options) || !('model_version' in analysis_options)) {
+      throw new Error('Setting analysis_type to "custom" requires additional params: "model_name" and "model_version"');
+    }
+    params.parameters = {
+      custom: {
+        model_name: analysis_options.model_name,
+        model_version: analysis_options.model_version
+      }
+    }
+  }
 
   let api_uri = ['analysis', 'analyze', 'uri'];
   return call_analysis_api('POST', api_uri, params, callback, options);
 }
 
 module.exports = {
-  analyze
+  analyze_uri
 };
--- a/lib/analysis/index.js
+++ b/lib/analysis/index.js
@@ -1,22 +1,20 @@
 const utils = require("../utils");
 const {call_analysis_api} = require('../api_client/call_analysis_api');
 
-function analyze_uri(uri, analysis_options, options = {}, callback) {
-  const analysisType = analysis_options.analysis_type;
-
+function analyze_uri(uri, analysis_type, options = {}, callback) {
   const params = {
     uri,
-    analysis_type: analysisType
+    analysis_type
   }
 
-  if (analysisType === 'custom') {
-    if (!('model_name' in analysis_options) || !('model_version' in analysis_options)) {
+  if (analysis_type === 'custom') {
+    if (!('model_name' in options) || !('model_version' in options)) {
       throw new Error('Setting analysis_type to "custom" requires additional params: "model_name" and "model_version"');
     }
     params.parameters = {
       custom: {
-        model_name: analysis_options.model_name,
-        model_version: analysis_options.model_version
+        model_name: options.model_name,
+        model_version: options.model_version
       }
     }
   }
--- a/lib/analysis/index.js
+++ b/lib/analysis/index.js
@@ -1,0 +1,17 @@
+const utils = require("../utils");
+const {call_analysis_api} = require('../api_client/call_analysis_api');
+
+function analyze(uri, analysis_type, options = {}, callback) {
+  let params = {
+    uri,
+    analysis_type,
+    parameters: options.analyze_parameters
+  };
+
+  let api_uri = ['analysis', 'analyze', 'uri'];
+  return call_analysis_api('POST', api_uri, params, callback, options);
+}
+
+module.exports = {
+  analyze
+};
--- a/lib/api_client/call_analysis_api.js
+++ b/lib/api_client/call_analysis_api.js
@@ -24,13 +24,7 @@ function call_analysis_api(method, uri, params, callback, options) {
   }
   options.content_type = 'json';
 
-  return execute_request(method, params, auth, api_url, callback, options)
-    .then(requestResult => {
-      return {
-        data: requestResult.data,
-        request_id: requestResult.request_id
-      };
-    });
+  return execute_request(method, params, auth, api_url, callback, options);
 }
 
 module.exports = {
--- a/lib/api_client/call_analysis_api.js
+++ b/lib/api_client/call_analysis_api.js
@@ -1,0 +1,38 @@
+const utils = require("../utils");
+const config = require("../config");
+const ensureOption = require('../utils/ensureOption').defaults(config());
+const execute_request = require("./execute_request");
+
+const {ensurePresenceOf} = utils;
+
+function call_analysis_api(method, uri, params, callback, options) {
+  ensurePresenceOf({
+    method,
+    uri
+  });
+  const api_url = utils.base_api_url_v2()(uri, options);
+  let auth = {};
+  if (options.oauth_token || config().oauth_token) {
+    auth = {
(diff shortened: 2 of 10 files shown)


lib/analysis/index.js
```diff
@@ -0,0 +1,17 @@
+const utils = require("../utils");
+const {call_analysis_api} = require('../api_client/call_analysis_api');
+
+function analyze(uri, analysis_type, options = {}, callback) {
```
adilester, line 4: Don't we want it to be called `analyze_uri`?
Soon we might have analyze by asset id or others.

lib/api_client/call_analysis_api.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+  }
+  options.content_type = 'json';
+
+  return execute_request(method, params, auth, api_url, callback, options)
+    .then(requestResult => {
+      return {
+        data: requestResult.data,
+        request_id: requestResult.request_id
+      };
```
adilester, line 32: Is it standard practice to only return selected fields?
If a new field is added to the API's response it won't be displayed.

lib/api_client/call_analysis_api.js
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    auth = {
+      key: ensureOption(options, "api_key"),
+      secret: ensureOption(options, "api_secret")
+    };
+  }
+  options.content_type = 'json';
+
+  return execute_request(method, params, auth, api_url, callback, options)
+    .then(requestResult => {
```
adilester, line 28: What happens on error?
We can have errors that look like this
```
{
  "error": {
    "category": "auth_error",
    "code": "MA_00010",
    "message": "authentication error",
    "details": {
      "message": "api_secret mismatch"
    }
  }
}
```

or we might have something else (unstructured) returned if something went really wrong.

lib/utils/index.js
```diff
@@ -1058,17 +1058,31 @@ function unsigned_url_prefix(
...
 
-function base_api_url(path = [], options = {}) {
-  let cloudinary = ensureOption(options, "upload_prefix", UPLOAD_PREFIX);
-  let cloud_name = ensureOption(options, "cloud_name");
-  let encode_path = unencoded_path => encodeURIComponent(unencoded_path).replace("'", '%27');
-  let encoded_path = Array.isArray(path) ? path.map(encode_path) : encode_path(path);
-  return [cloudinary, "v1_1", cloud_name].concat(encoded_path).join("/");
+function base_api_url_v1() {
+  return base_api_url('v1_1');
+}
```
adilester, line 1063: Maybe rename to `base_api_url_v1_1`?

lib/utils/index.js
```diff
@@ -1058,17 +1058,31 @@ function unsigned_url_prefix(
...
+function base_api_url_v2() {
+  return base_api_url('v2');
+}
+
+function base_api_url(api_version) {
+  if (!api_version || api_version.length === 0) {
+    throw new Error('api_version needs to be a non-empty string');
+  }
+
+  return (path = [], options = []) => {
```
adilester, line 1074: This looks kinda weird to me. Is this idiomatic to node?

lib/utils/index.js
```diff
@@ -1058,17 +1058,31 @@ function unsigned_url_prefix(
...
-  let encode_path = unencoded_path => encodeURIComponent(unencoded_path).replace("'", '%27');
-  let encoded_path = Array.isArray(path) ? path.map(encode_path) : encode_path(path);
-  return [cloudinary, "v1_1", cloud_name].concat(encoded_path).join("/");
+function base_api_url_v1() {
+  return base_api_url('v1_1');
+}
+
+function base_api_url_v2() {
+  return base_api_url('v2');
+}
```
adilester, line 1067: Maybe define these as consts instead of functions? Then the call wouldn't look weird with two function invocations (`func()(params)`)

lib/utils/index.js
```diff
@@ -1058,17 +1058,31 @@ function unsigned_url_prefix(
...
+
+function base_api_url(api_version) {
```
adilester, line 1069: `get_base_api_url_func`?

types/cloudinary_ts_spec.ts
```diff
@@ -1148,6 +1148,15 @@ cloudinary.v2.provisioning.account.user_group_users(
...
 
+// $ExpectType Promise<AnalyzeResponse>
+cloudinary.v2.analysis.analyze('https://example.com', 'custom', {
+    analyze_parameters: {
+        custom: {
+            model_name: 'my_model',
+            model_version: 1
+        }
+    }
+});
```
adilester, line 1159: What does this do?

types/index.d.ts
```diff
@@ -836,6 +836,25 @@ declare module 'cloudinary' {
         value: string;
     }
 
+    export type AnalysisType = 'captioning' | 'cld_fashion' | 'cld_text' | 'coco' | 'google_tagging' | 'human_anatomy' | 'lvis' | 'shop_classifier' | 'unidet' | 'custom';
+
+    export interface AnalyzeResponse {
+        data: {
+            entity: string,
+            analysis: Record<string, Record<string, string> | Array<string> | string>
```
adilester, line 844: Is this generally how custom JSON is represented?

types/index.d.ts
```diff
@@ -836,6 +836,25 @@ declare module 'cloudinary' {
...
+        },
+        request_id: string,
+    }
+
+    export interface AnalyzeOptions {
+        analyze_parameters: {
```
adilester, line 850: Shouldn't it be just `parameters`? or does it not matter?

lib/api_client/call_analysis_api.js
cloudinary-pkoniu: It'll be returned correctly - I verified by sending incorrect `analysis_type` and credentials and got two different errors but looking exactly like you sent.

lib/analysis/index.js
cloudinary-pkoniu: Good idea, changed the name.

lib/utils/index.js
cloudinary-pkoniu: It's a very common pattern called function currying: https://javascript.info/currying-partials

types/cloudinary_ts_spec.ts
cloudinary-pkoniu: If you take a look at the comment one line above - that's the assertion. It verifies if `analyze_uri` returns a Promise with `AnalyzeResponse`

types/index.d.ts
cloudinary-pkoniu: I changed the interface a little bit

lib/utils/index.js
cloudinary-pkoniu: I wanted to keep changes here to bare minimum - it was a function earlier, so I wanted to keep it that way for now.

types/index.d.ts
cloudinary-pkoniu: Yes, `Record<>` is object, later on you specify the type for keys and values. So in this case `analysis` can be an object with strings as keys and another object as value or an array with strings or just a string.

types/index.d.ts
```diff
@@ -836,7 +836,23 @@ declare module 'cloudinary' {
         value: string;
     }
 
-    export type AnalysisType = 'captioning' | 'cld_fashion' | 'cld_text' | 'coco' | 'google_tagging' | 'human_anatomy' | 'lvis' | 'shop_classifier' | 'unidet' | 'custom';
+    export type AnalysisType = CustomizableAnalysisType | NonCustomizableAnalysisType;
+
+    export type CustomizableAnalysisType = 'custom';
```
adilester, line 841: Please assume that all analysis types will be customizable (i.e. accept parameters)
It's just not the case right now because the API is not very mature.

Also, as mentioned before, some parameters will have the same name. For example, soon `model_version` can be part of `captioning` as well, so both of these will be valid:

```
{
	"analysis_type": "captioning",
    "parameters":
    {
        "captioning":
        {
            "model_version": 4
        }
    }
}
```

```
{
	"analysis_type": "custom",
    "parameters":
    {
        "custom":
        {
        	"model_name": "my-model",
            "model_version": 4
        }
    }
}
```

const-cloudinary (approved):

lib/utils/index.js
pawelphilipczyk-cloudinary: Same thought I had, if we have a specific version here and we put the version in the name then I guess it makes sense to be specific in the name too.

pawelphilipczyk-cloudinary (approved):

PR #659, opened 2024-04-12
feat: selective response for admin and search api

cloudinary-pkoniu:
### Brief Summary of Changes
Support selective response for Admin and Search API: `fields` parameter is correctly added to requests supporting this parameter.

- [ ] Refactoring
- [X] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -71,7 +71,7 @@ exports.resources_by_moderation = function resources_by_moderation(kind, status,
   let resource_type, uri;
   resource_type = options.resource_type || "image";
   uri = ["resources", resource_type, "moderations", kind, status];
-  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "tags", "context", "direction", "moderations", "metadata", "fields", "fields"), callback, options);
+  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "tags", "context", "direction", "moderations", "metadata", "fields"), callback, options);
 };
 
 exports.resource_by_asset_id = function resource_by_asset_id(asset_id, callback, options = {}) {
--- a/lib/api.js
+++ b/lib/api.js
@@ -45,21 +45,21 @@ exports.resources = function resources(callback, options = {}) {
   if ((options.start_at != null) && Object.prototype.toString.call(options.start_at) === '[object Date]') {
     options.start_at = options.start_at.toUTCString();
   }
-  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "prefix", "tags", "context", "direction", "moderations", "start_at", "metadata"), callback, options);
+  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "prefix", "tags", "context", "direction", "moderations", "start_at", "metadata", "fields"), callback, options);
 };
 
 exports.resources_by_tag = function resources_by_tag(tag, callback, options = {}) {
   let resource_type, uri;
   resource_type = options.resource_type || "image";
   uri = ["resources", resource_type, "tags", tag];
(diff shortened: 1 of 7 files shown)


lib/api.js
```diff
@@ -71,7 +71,7 @@ exports.resources_by_moderation = function resources_by_moderation(kind, status,
   let resource_type, uri;
   resource_type = options.resource_type || "image";
   uri = ["resources", resource_type, "moderations", kind, status];
-  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "tags", "context", "direction", "moderations", "metadata"), callback, options);
+  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "tags", "context", "direction", "moderations", "metadata", "fields", "fields"), callback, options);
```
const-cloudinary, line 74: Maybe 
```suggestion
  return call_api("get", uri, pickOnlyExistingValues(options, "next_cursor", "max_results", "tags", "context", "direction", "moderations", "metadata", "fields"), callback, options);
```

?

const-cloudinary (changes requested):

lib/api.js
cloudinary-pkoniu: You're right, copy-paste mistake it seems. Fixed.

const-cloudinary (approved):

test/integration/api/admin/api_spec.js
```diff
@@ -17,6 +17,8 @@ const retry = require('../../../testUtils/helpers/retry');
 const {shouldTestFeature} = require("../../../spechelper");
 const API_V2 = cloudinary.v2.api;
 const DYNAMIC_FOLDERS = helper.DYNAMIC_FOLDERS;
+const assert = require('assert');
```
pawelphilipczyk-cloudinary, line 20: `require` is the same as `import` to me and usually those go on the top of the file. Not a strong opinion though because `require` is written with variable assignment anyway.

test/integration/api/admin/api_spec.js
```diff
@@ -425,6 +427,51 @@ describe("api", function () {
         }));
       });
     });
+
+    describe('selective response', () => {
+      const expectedKeys = ['public_id', 'asset_id', 'folder', 'tags'].sort();
```
pawelphilipczyk-cloudinary, line 432: Maybe add a comment here on why stating like this and sorting instead of defining in a sorted order upfront?

test/integration/api/admin/api_spec.js
```diff
@@ -425,6 +427,51 @@ describe("api", function () {
...
+
+      it('should allow listing', async () => {
+        const {resources} = await cloudinary.v2.api.resources({fields: ['tags']})
+        const actualKeys = Object.keys(resources[0]);
```
pawelphilipczyk-cloudinary, line 436: There is a lot of repetition here so worth considering adding a helper like `assertKeysEqual(resource[0])` which would do all the `Object.keys` and `deepStrictEqual` and `sort`ing inside of it. 

That would probably make the tests more readable too. Not that they are somehow very complicated, just a thought.

lib/v2/search.js
```diff
@@ -91,6 +96,16 @@ const Search = class Search {
     return this;
   }
 
+  fields(value) {
```
pawelphilipczyk-cloudinary, line 99: It's possible to get a unique set of values from an array by using `Array.from(new Set(this.query_hash_fields))`.

I know it's the same in more places here so just a hint that maybe it would reduce the boilerplate of having each field doing this `find`ing and `push`ing.

pawelphilipczyk-cloudinary (approved):

lib/v2/search.js
cloudinary-pkoniu: Thanks for the suggestion! Changed the implementation accordingly.

PR #674, opened 2024-07-19
fix: use 0.0.0 as fallback when package.json unavailable

cloudinary-pkoniu:
### Brief Summary of Changes
Under certain circumstances when package.json is unavailable, the sdk could not read the version of itself (which is necessary to calculate analytics token appended to the url). It caused an error to be thrown "Must provide sdkSemver". This PR fixes it and allows sdk to fail gracefully - `0.0.0` is used as a fallback if package.json is unavailable.

- [ ] Refactoring
- [ ] New feature
- [X] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [X] Yes
- [ ] No

### DIFF
--- a/lib/utils/analytics/getSDKVersions.js
+++ b/lib/utils/analytics/getSDKVersions.js
@@ -2,6 +2,19 @@ const fs = require('fs');
 const path = require('path');
 const sdkCode = 'M'; // Constant per SDK
 
+function readSdkSemver() {
+  const pkgJsonPath = path.join(__dirname, '../../../package.json');
+  try {
+    const pkgJSONFile = fs.readFileSync(pkgJsonPath, 'utf-8');
+    return JSON.parse(pkgJSONFile).version
+  } catch (e) {
+    if (e.code === 'ENOENT') {
+      return '0.0.0'
+    }
+    return 'n/a';
+  }
+}
+
 /**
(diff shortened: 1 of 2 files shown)


lib/utils/analytics/getSDKVersions.js
```diff
@@ -2,17 +2,28 @@ const fs = require('fs');
...
+function readSdkSemver() {
+  const pkgJsonPath = path.join(__dirname, '../../../package.json');
+  try {
+    const pkgJSONFile = fs.readFileSync(pkgJsonPath, 'utf-8');
+    return JSON.parse(pkgJSONFile).version
+  } catch (e) {
+    if (e.code === 'ENOENT') {
+      return '0.0.0'
+    }
+    return 'n/a';
```
const-cloudinary, line 14: Is there a test for `n/a`? I would change it to something we can decode or would not cause additional errors.

const-cloudinary (changes requested):

lib/utils/analytics/getSDKVersions.js
cloudinary-pkoniu: Yes, there is - the resulting token will be `_a=E` because it assumes numerical values

const-cloudinary (approved):

Issue #675, opened 2024-07-23
Missing download_folder method


vin-krish:
## Describe the bug in a sentence or two.

documentation here shows download_folder method available in nodejs https://cloudinary.com/documentation/image_upload_api_reference#download_folder_syntax 

After installing multiple version in 1.x.x and 2.x.x, couldn't find this download_folder method at all.

version: "cloudinary": "^2.3.0"

## Issue Type (Can be multiple)
[x] Behaviour - Functions aren’t working as expected (Such as generate URL)
[x] Documentation - Inconsistency between the docs and behaviour

## Steps to reproduce
install the version and try to find the method.

## Browsers (if issue relates to UI, else ignore)
[ ] Chrome
[ ] Firefox
[ ] Safari
[ ] Other (Specify)
[ ] All

## Versions and Libraries (fill in the version numbers)
Cloudinary_NPM SDK version
Node - v18.12.1
NPM - 8.19.2

## Config Files (Please paste the following files if possible)
Package.json

## Repository
If possible, please provide a link to a reproducible repository that showcases the problem

vin-krish:
@tamaracloudinary found the issue, this method is not exposed in typescript types:
https://github.com/cloudinary/cloudinary_npm/blob/master/types/index.d.ts#L925

hence wasn't able to call from my TS code.

PR #677, opened 2024-07-24
Add missing type definition

vin-krish:
### Brief Summary of Changes
Typescript type definitions doesn't include all the methods, notably download_folder which I was trying to use instead of download_zip_url because this has a limit of 1000 asset.

- [ ] GitHub issue (Add reference - #675)

#### Are Tests Included?
- [x] Yes
- [ ] No

#### Reviewer, Please Note:
No new code added, this just adds an existing method in js to typescript type definition.

### DIFF
--- a/types/cloudinary_ts_spec.ts
+++ b/types/cloudinary_ts_spec.ts
@@ -916,7 +916,7 @@ cloudinary.v2.utils.download_folder(
     'my-folder',
     {
         resource_type: 'image',
-        prefix: 'photos/'
+        prefixes: 'photos/'
     }
 );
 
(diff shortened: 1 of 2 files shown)


types/cloudinary_ts_spec.ts
```diff
@@ -911,6 +911,16 @@ cloudinary.v2.utils.download_zip_url(
     }
 );
 
+// $ExpectType string
+cloudinary.v2.utils.download_folder(
+    'my-folder',
+    {
+        resource_type: 'image',
+        prefix: 'photos/'
```
cloudinary-pkoniu, line 919:
```suggestion
        prefixes: 'photos/'
```

cloudinary-pkoniu (changes requested):
Small typo in ts_spec, other than that, good work!

types/cloudinary_ts_spec.ts
vin-krish: thank you, fixed it.

PR #707, opened 2025-04-21
Feature: Add support for restoring asset(s) by asset id

esguerrat:
### Brief Summary of Changes

This PR adds support for restoring one or more assets by asset_id, via the following endpoint: POST /resources/restore/:asset_ids
Documentation reference: https://cloudinary.com/documentation/admin_api#restore_resources_by_asset_id

- [ ] Refactoring
- [ x] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ x] Yes
- [ ] No


types/index.d.ts
```diff
@@ -1102,6 +1102,10 @@ declare module 'cloudinary' {
 
             function restore(public_ids: string[], callback?: ResponseCallback): Promise<any>;
 
+             function restore_by_asset_ids(asset_ids: string[] | string, options?: AdminAndResourceOptions, callback?: ResponseCallback): Promise<ResourceApiResponse>;
```
cloudinary-pkoniu, line 1105: If I see correctly, the implementation of `restore_by_asset_ids` doesn't support passing a single value, we're only passing the first argument as a payload when doing the request to Admin API.
Currently Admin API only supports a list of values under `asset_ids` which means that, either this type definition needs to change to `string[]` or we should wrap the first argument in array if it's not already an array, in the `restore_by_asset_ids` implementation.

package.json
```diff
@@ -29,7 +29,7 @@
     "jsdoc": "^4.0.4",
     "jsdom": "^9.12.0",
     "jsdom-global": "2.1.1",
-    "mocha": "^6.2.3",
+    "mocha": "^7.2.0",
```
cloudinary-pkoniu, line 32: Was this intentional?

cloudinary-pkoniu (changes requested):
Great job, I left two comments, please take a look.

Also, if I see correctly, some specs are failing in some of the node versions, please take a look at those too.

types/index.d.ts
```diff
@@ -1102,6 +1102,8 @@ declare module 'cloudinary' {
 
             function restore(public_ids: string[], callback?: ResponseCallback): Promise<any>;
 
+            function restore_by_asset_ids(asset_ids: string[] | string, options?: AdminAndResourceOptions, callback?: ResponseCallback): Promise<ResourceApiResponse>;
```
cloudinary-pkoniu, line 1105:
```suggestion
            function restore_by_asset_ids(asset_ids: string[], options?: AdminAndResourceOptions, callback?: ResponseCallback): Promise<ResourceApiResponse>;
```

PR #711, opened 2025-06-28
chore: replace all instances of q with native promise

IshfaqAhmedProg:
#### What Does This PR Address?
- [x] GitHub issue (Add reference - #686)
- [ ] Refactoring
- [ ] New feature
- [ ] Bug fix
- [ ] Adds more tests

#### Are Tests Included?
- [ ] Yes
- [x] No

#### Reviewer, Please Note:

On uploader_spec.js 806:14 and 830:14 i replaced Q.denodeify directly with the function call i dont know why it wasn't implemented like that before hand, so please take care to look into it.

PR #720, opened 2025-10-28
refactor: use native promises and drop q

### LINKED ISSUE #686: Q dependency is deprecated
## Bug report for Cloudinary NPM SDK
Before proceeding, please update to latest version and test if the issue persists

## Describe the bug in a sentence or two.
I get this warning on every build:

```
npm warn deprecated q@1.5.1: You or someone you depend on is using Q, the JavaScript Promise library that gave JavaScript developers strong feelings about promises. They can almost certainly migrate to the native JavaScript promise now. Thank you literally everyone for joining me in this bet against the odds. Be excellent to each other.
npm warn deprecated
npm warn deprecated (For a CapTP with native promises, see @endo/eventual-send and @endo/captp)
```

Is there a reason why native promises aren't used yet?

## Issue Type (Can be multiple)
[ ] Build - Can’t install or import the SDK
[ ] Babel - Babel errors or cross browser issues
[x] Performance - Performance issues
[ ] Behaviour - Functions aren’t working as expected (Such as generate URL)
[ ] Documentation - Inconsistency between the docs and behaviour
[ ] Incorrect Types - For typescript users who are having problems with our d.ts files
[x] Other (Specify)

## Versions and Libraries (fill in the version numbers)
Cloudinary_NPM SDK version 2.4.0
Node - 20.17.0
NPM - 10.8.2

AviVahl:
migrated away from q, and used native promises.

unhandled rejections in unit tests now cause the test runner to fail. ensured all rejections are ignored with no-op fn.

found and fixed several broken tests that did not report failures correctly.

fixed wait() not used correctly outside .then, as it's returning a higher order fn rather than the promise (to chain values in .then callbacks)

implemented a tiny test polyfill for Promise.allSettled to support older Node versions.

worked around a timezone issue and a test failing  when running it in ~1am locally, heh.

added ability to use .only/skip on the test proxy of mocha's describe fn

closes #711 
closes #686

### DIFF
--- a/test/testUtils/suite.js
+++ b/test/testUtils/suite.js
@@ -12,4 +12,7 @@ function makeSuite(name, tests) {
   });
 }
 
+makeSuite.only = describe.only;
+makeSuite.skip = describe.skip;
+
 module.exports = makeSuite;
--- a/lib/api_client/execute_request.js
+++ b/lib/api_client/execute_request.js
@@ -2,7 +2,6 @@
 const config = require("../config");
 const https = /^http:/.test(config().upload_prefix) ? require('http') : require('https');
 const querystring = require("querystring");
-const Q = require('q');
 const url = require('url');
 const utils = require("../utils");
 const ensureOption = require('../utils/ensureOption').defaults(config());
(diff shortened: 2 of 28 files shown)


test/testUtils/suite.js
```diff
@@ -12,4 +12,7 @@ function makeSuite(name, tests) {
   });
 }
 
+makeSuite.only = describe.only;
```
AviVahl: can now .only/skip on describe calls

AviVahl:
any chance for a review? I feel like this one gets this SDK one step closer to 2025.

